PR #1, opened 2017-03-13
PCHR-2064: Merge missing changes done in CiviHR

AkA84:
This PR merges in the shoreditch theme the changes that had been made to the "local" civihr theme since shoreditch had been moved to a different repo

Additionally, a SASS interpolation had been fixed that solves wrong styling on the "Panels With Tabs - Horizontal Tabs" component

*Before*
```scss
.panel-subheading {
  > .nav-tabs {
    margin: -#{$padding-y - 5} -#{$padding-x} -#{$padding-y};
  }
}
```
```css
.panel-subheading > .nav-tabs { margin: -10px - 20px - 15px; }
```
panel-tabs-before

*After*
```scss
.panel-subheading {
  > .nav-tabs {
    margin: -#{$padding-y - 5} #{-$padding-x} #{-$padding-y};
  }
}
```
```css
.panel-subheading > .nav-tabs { margin: -10px -20px -15px; }
```
panel-tabs-after

---

Another fix was done related to the `.required` class that was added to the theme. This was conflicting with the old `civicrm-*` markup which was using also a `.required` class

required-before

So I changed the class name to `.required-mark` (as in, the "*" to mark required fields) so that it doesn't conflict with `.required` which is in fact a too-common class name

required-after

PR #4, opened 2017-03-30
PCHR-2064 merge civihr changes

AkA84:
PR #1 has been erroneously merged before being tested internally, so it had to be rolled back

# Update
This PR, compared to the previous one, contains additional fixes for dropdowns and input fields on standard civicrm pages

It also slightly improves the gulpfile, by using the common `task:sub-task` notation, and by returning the sass tasks only when they're done


AkA84:
@totten this pr will be ready to go tomorrow morning together with others on the civihr repo, please do not merge

gulpfile.js
```diff
@@ -27,8 +27,8 @@ gulp.task('sass-bootstrap', ['sass-sync'], function () {
     .pipe(gulp.dest('css/'));
 });
 
-gulp.task('sass-civicrm', ['sass-sync'], function () {
-  gulp.src('scss/civicrm/custom-civicrm.scss')
+gulp.task('sass:civicrm', ['sass:sync'], function () {
+  return gulp.src('scss/civicrm/custom-civicrm.scss')
```
totten: Ah, nice. So this is probably why `gulp` was hanging for a moment before finishing?

https://github.com/gulpjs/gulp/blob/master/docs/API.md#async-task-support

AkA84: i think the hanging was due to how massive the custom-civicrm.css file is (it relies on way to  many `@extends` at the moment).

What `return` does is to allow gulp to understand when the task has effectively done running, otherwise without it you would get something like this
```js
[08:15:33] Starting 'sass:bootstrap'...
[08:15:33] Finished 'sass:bootstrap' after 13 ms
[08:15:33] Starting 'sass:civicrm'...
[08:15:33] Finished 'sass:civicrm' after 2.8 ms
[08:15:33] Starting 'sass'...
[08:15:33] Finished 'sass' after 2.81 μs
```
instead of this
```js
[08:14:41] Starting 'sass:bootstrap'...
[08:14:41] Starting 'sass:civicrm'...
[08:14:50] Finished 'sass:bootstrap' after 8.79 s
[08:15:03] Finished 'sass:civicrm' after 22 s
[08:15:03] Starting 'sass'...
[08:15:03] Finished 'sass' after 9.94 μs
```
notice that in the first case gulp reports `sass:civicrm` to be done after 3 ms, but in reality sass is still processing the files

PR #5, opened 2017-03-31
General Styling Improvements 

meladawy:
This contain a general styling improvements for shoreditsh extension which include the following 

1- Defining a new Wizard Component

Wizard Component HTMl Structure 

```html
<div class="crm_wizard">

    <div class="panel panel-default crm_wizard__title">
        <div class="panel-body">
            <ul class="nav nav-pills">
                <li class="active">
                    <a href="">
                        <span class="crm_wizard__title__number"> 1 </span>
                        Step 1
                    </a>
                </li>
                
                <li>
                    <a href="">
                        <span class="crm_wizard__title__number"> 2 </span> 
                        Step 2
                    </a>
                </li>
                
                <li>
                    <a href="">
                        <span class="crm_wizard__title__number"> 3 </span> 
                        Step 3
                    </a>
                </li>
            </ul>
        </div>
    </div>

    <div class="panel panel-default crm_wizard__body">
        <div class="panel-body">
            Any body content
        </div>
    </div>

    <div class="panel panel-default crm_wizard__footer">
        <div class="panel-body">
            <div class="row">
                <div class="col-md-6 text-left">
                    <button class="btn-danger btn">Action</button>
                </div>
                <div class="col-md-6 text-right">
                    <button class="btn-primary btn">Action</button>
                </div>
            </div>
        </div>
    </div>

</div>
```

Wizard Component Output

screen shot 2017-03-30 at 3 15 16 pm

2- Add prefix mixin that can be used across the application
3- Add shadows in different website panels 

screen shot 2017-03-31 at 2 19 46 pm

4- Improvements in select2 widget design 
5- Define a new class ".form-inline-tabs" that can be used with form-group as following 

```html
<div class="form-inline form-inline-tabs">
    <div class="form-group">
        <div class="radio">
          <input type="radio" name="send" value="now">
          <label>Send immediately</label>
        </div>
    </div>
    <div class="form-group active">
        <div class="radio">
          <input type="radio" name="send" value="at" checked="checked">
          <label>Send on..</label>
        </div>
    </div>
</div>
```

and output 

screen shot 2017-03-31 at 2 23 46 pm

6- Change the placeholder color across text fields and make it lighter


scss/bootstrap/overrides/style/_forms.scss
```diff
@@ -75,6 +89,36 @@ label {
   }
 }
 
+// Style new group of type radio
+.form-inline {
```
AkA84: I don't think this is necessary, you can simply do

```scss
.form-inline-tabs {

}
```
in bootstrap class variations are always applied together with the base classe, like
```scss
<button class="btn btn-primary"></button>
```
so you don't have the selector to be
```scss
.form-inline.form-inline-tabs
```

scss/bootstrap/overrides/style/_forms.scss
```diff
@@ -75,6 +89,36 @@ label {
...
+
+  &.form-inline-tabs .form-group {
```
AkA84: always use nesting
```scss
.form-inline-tabs {

  .form-group {

  }
}
```

scss/bootstrap/overrides/style/_forms.scss
```diff
@@ -75,6 +89,36 @@ label {
...
+    box-shadow: $crm-form-control-shadow;
+    border: 1px solid $crm-darker-grayblue;
+    line-height: 1em;
+    padding: 8px 15px;
+    margin-right: -5px;
+    min-width: 180px;
```
AkA84: list the properties alphabetically

scss/bootstrap/overrides/style/_forms.scss
```diff
@@ -75,6 +89,36 @@ label {
...
 
+// Style new group of type radio
+.form-inline {
+
+  &.form-inline-tabs .form-group {
+    box-shadow: $crm-form-control-shadow;
+    border: 1px solid $crm-darker-grayblue;
+    line-height: 1em;
+    padding: 8px 15px;
+    margin-right: -5px;
```
AkA84: what is this margin-right for? this look like a "magic number" (arbitrary number to just make things work), where does this value come from?

scss/bootstrap/overrides/style/_forms.scss
```diff
@@ -75,6 +89,36 @@ label {
...
+    min-width: 180px;
+
+
+    &.active {
+      background-color: $gray-lighter;
+    }
+
+    &:first-child {
+      border-radius: 2px 0 0 2px;
```
AkA84: you already have the border radius of 2px defined in a bootstrap variable, please use that

scss/bootstrap/overrides/style/_forms.scss
```diff
@@ -75,6 +89,36 @@ label {
...
+    }
+
+    &:last-child {
+      border-radius: 0 2px 2px 0;
+    }
+
+    input {
+      margin-right: 10px;
```
AkA84: why is this necessary?

scss/bootstrap/overrides/style/_forms.scss
```diff
@@ -75,6 +89,36 @@ label {
   }
 }
 
+// Style new group of type radio
```
AkA84: Define briefly what this style is for

scss/bootstrap/overrides/style/_panel.scss
```diff
@@ -1,6 +1,4 @@
 .panel {
-  @include box-shadow(none);
```
AkA84: put this back please

scss/bootstrap/overrides/style/_forms.scss
AkA84: Also, this is a variation that needs to be covered in the styleguide

scss/bootstrap/overrides/style/_forms.scss
meladawy: Yea i just followed bootstrap way of defining inline form groups to save some extra styling

http://getbootstrap.com/css/#forms-inline

so should i remove form-inline and add the inline styling under "form-inline-tabs" ?
```
.form-inline-tabs .form-group {
  display: inline-block;
}
```

scss/bootstrap/overrides/style/_forms.scss
meladawy: I did this to avoid having 4 levels which is not recommended in SASS

```
.form-inline {
  &.form-inline-tabs {
    .form-group {
      box-shadow: $crm-form-control-shadow;
      border: 1px solid $crm-darker-grayblue;
      line-height: 1em;
      padding: 8px 15px;
      margin-right: -5px;
      min-width: 180px;

      &.active {
        background-color: $gray-lighter;
      }

      &:first-child {
        border-radius: 2px 0 0 2px;
      }

      &:last-child {
        border-radius: 0 2px 2px 0;
      }

      input {
        margin-right: 10px;
      }
    }
  }
}
```

scss/bootstrap/overrides/style/_forms.scss
meladawy: This is to remove the spacing between elements. Without marging right it would be like this 

screen shot 2017-04-06 at 11 05 15 am

Between the target to be like this 

screen shot 2017-04-06 at 11 05 37 am


So giving the groups a float:left; OR margin-right:-5px should resolve this

scss/bootstrap/overrides/style/_forms.scss
meladawy: I updated this to be like 
```
    input[type="radio"] {
      margin-right: 10px;
    }
```

Radio inputs require some right margin 

sdasd

scss/bootstrap/overrides/style/_forms.scss
AkA84: yes
```scss
.form-inline-tabs {
  // everything else here
}

scss/bootstrap/overrides/style/_forms.scss
AkA84: since you don't need to use `.form-inline` in the selector anymore, you have 3 levels

Anyway it is a matter of consistent, it's easier to scan and read the file if you know that selectors are always nested, instead of having a mix of nested and inline

scss/bootstrap/overrides/style/_forms.scss
AkA84: Does that margin come from somewhere (as in, was it inherited)? or is it something that happen because the elements are inline?

any solution that relies either on a non-numeric value property (like `float`) or that it uses a value that comes from a shared variable (say something like `margin-right: -#{$some-common-margin-value}`) is preferrable over a solution with arbitrary numbers

scss/bootstrap/overrides/style/_forms.scss
AkA84: But radio buttons already have, by default, a right margin (see [here](http://getbootstrap.com/css/#checkboxes-and-radios)), so this should not be necessary

meladawy: It doesn't has in the example. but it has margin-left:-20px; 
and it also should has the wrapping ".radio" to take this margin-left:-20px; so i have to change the markup to match this 

I also tried to change the markup from inspect element and i found that you override the margin-left:-20px  in shoreditsh theme

screen shot 2017-04-06 at 11 45 17 am

scss/bootstrap/overrides/style/_forms.scss
meladawy: Its only coming from inline. no margin in the fields. There is not common margin variable for this. Should i replace this with float left?

meladawy:
@AkA84 I resolved the comments except one comment for radio buttons margin right. Can you please review again and let me know if you still have any concerns.

guanhuan:
Styling looks great to me. @AkA84 can you confirm you are happy with the comments are resolved.

PR #9, opened 2017-04-27
HW-293: Updating FontAwesome to 4.7

meladawy:
Upgrading FontAwesome to 4.7 to get all icons working  

Before: 
screen shot 2017-04-27 at 2 37 35 pm

After:
screen shot 2017-04-28 at 1 36 26 pm

BackstopJs: 
screen shot 2017-04-28 at 1 34 03 pm


meladawy:
References:- 

Most of references refer to (Cross domain access issue) between i can confirm that font is loaded from the same domain (no CDN) and some icons working properly so its FontAwesome is loaded successfully so its not "Access to origin" issue 

1- http://stackoverflow.com/questions/16866872/font-awesome-not-working-in-firefox
2- https://github.com/cdnjs/cdnjs/issues/755
3- https://support.mozilla.org/en-US/questions/1003474


After some discussion with @bmnepali, we found that we are using FontAwesome 4.4.0 and "\f2d0" (fa-window-maximize) is not exist in version 4.4.0. 

https://github.com/FortAwesome/Font-Awesome/blob/v4.4.0/less/variables.less

So maybe is not loaded in Firefox because this icon is not exist in this FontAwesome version but the question is. Why i can see the icon from Chrome then?

bmnepali:
@AkA84 and @meladawy  Form within `uk.co.compucorp.civicrm.hrcore`, I ran `gulp backstopjs:reference` task checking out to all 6 civihr repos to `staging` branch and ran `gulp backstopjs:test` checking out to `HW-293-browser-compatability-issues` in `org.civicrm.shoreditch` extension,  then i have following result:
screen shot 2017-04-28 at 9 15 44 pm

Since, sometimes the datepicker dialog gets lacked, it should not be the problem to the changes introduces by this PR. So, i am approving it.

PR #34, opened 2017-08-21
HW-373:Fixed editable content section globally

mukeshcompucorp:
**Before:**
before

**After:**
**_On hover:_**
_textarea:_
after

_text:_
before-2

_dropdown (select/boolean):_
before-2

**_On Edit:_**
_textarea:_
after-2

_text:_
after-3

_dropdown (select/boolean):_
after-4

**Note:**

1. This commit resolves design for all editable sections across the shoreditch extension 
2.  "Subject:" is representative label for civicase not for shoreditch editable content
3. This does not design custom fields group section editing portion as it is having civicase specific changes.


scss/bootstrap/components/_editable-content.scss
```diff
@@ -0,0 +1,31 @@
...
+    content: "";
+    height: 100%;
+    position: absolute;
+    right: 0;
+    text-align: center;
+    vertical-align: middle;
+    top: 0;
+    width: 30px;
+  }
+  &:after {
```
meladawy: Can u leave an empty line between selectors?

scss/bootstrap/components/_editable-content.scss
```diff
@@ -0,0 +1,31 @@
+.crm-editable-enabled:not(.crm-editable-editing):hover {
```
AkA84: where does this class come from? what are class editable contents have?

mukeshcompucorp: @AkA84 crm-editable-enabled comes from modules/civicrm/js/jquery/jquery.crmEditable.js and is generated for a section which is having enabled editing.

mukeshcompucorp: @AkA84 , should we keep pencil edit on data-type boolean or select,
When an editable content gets converted to dropdown box?

test-1-hw-373
Transforms to,
test-2-hw-373

I have checked other changes and fixed it. Let me know before we commit to this.

mukeshcompucorp:
@guanhuan In-Place editor is not styled as per design by Jon.
after-2

**Note:** 
Changes specific to custom field action buttons(save/cancel) are not in this PR as it seems civicase specific.
I will have a separate PR for that style and Subject label.

scss/bootstrap/components/_editable-content.scss
```diff
@@ -0,0 +1,85 @@
...
+      height: 100%;
+      position: absolute;
+      right: 0;
+      text-align: center;
+      top: 0;
+      vertical-align: middle;
+      width: 30px;
+    }
+
+    &:after {
```
AkA84: I don't think you need this, i think font awesome already provides some mixin? Please check in vendor/font-awesome

mukeshcompucorp: @AkA84 Fixed this.

scss/bootstrap/overrides/_variables.scss
```diff
@@ -83,6 +83,7 @@ $body-bg: $crm-background !default;
 
 $font-family-sans-serif:  "Helvetica Neue", Helvetica, Arial, sans-serif !default;
 $font-family-base:        "Open Sans", $font-family-sans-serif !default;
+$font-family-fontawesome: 'FontAwesome';
```
AkA84: please remove this, you're not using it

mukeshcompucorp: @AkA84 This is fixed

PR #37, opened 2017-08-31
AI-150:Fixed modal buttons regression issue

mukeshcompucorp:
Main part of this issues is fixed at https://github.com/compucorp/org.civicrm.shoreditch/pull/28

This PR fixes 
- margin for .fa-icon 
- hover/active effect of button

**Before:**
before-4

**After:**
after-4

**Note:**
This PR resolves button issue in popup throughout CiviCRM, backstopjs tests passed.


AkA84:
@mukeshcompucorp if this is a regression issue, please explain where the regression happen in the PR description

AkA84:
@mukeshcompucorp also please check this PR https://github.com/compucorp/org.civicrm.shoreditch/pull/28 , because it seems to me that the issue is the same

scss/civicrm/common/_modals.scss
```diff
@@ -141,15 +141,23 @@
           margin-left: 0 !important;
         }
 
-        &:hover {
+        &:hover,
+        &:active,
+        &:focus {
           color: $crm-white;
-          opacity: .8;
+          opacity: .8 !important;
```
AkA84: why the !important?

mukeshcompucorp: the button extends the %btn-civi which has opacity: 1 !important, to override that we need important here.

PR #45, opened 2017-09-05
AI-144:Fixed Find Participant Styling

mukeshcompucorp:
**Before:**
before-1
before-2

**After:**
after-1

## Comments

- .crm-submit-button is styled globally and removed from other files
- Sdditionally removes vacancies module of CiviHR and adds in CiviHR repo with PR https://github.com/civicrm/civihr/pull/2162


scss/civicrm/contact/pages/_events.scss
```diff
@@ -15,6 +15,34 @@
     }
   }
 
+  #searchForm {
+    @include border-box();
+    padding: 20px;
+  }
+
+  .crm-form-submit {
```
AkA84: aren't we styling the buttons already? Why is this button not getting the style? Is the markup different? If it is, don't we have some sass mixing that we can `@include` instead of having to re-define all the values again?

scss/civicrm/common/_buttons.scss
```diff
@@ -59,6 +59,52 @@ $button-border-radius: 2px;
     }
   }
 
+  input.crm-form-submit {
```
AkA84: do you need to specify `input`? Can't it be just `.crm-form-submit` ?

mukeshcompucorp: We need this as civicrm.css from main theme has default style with same input.crm-form-submit and that prevents us from overriding style if that's just `.crm-form-submit`

PR #48, opened 2017-09-08
HW-381: Styled case type settings screen

mukeshcompucorp:
**Before:**
before

**After:**
after-2

**Notes:**
- This PR creates common style for .crm-submit-buttons (which works as button panel) to get common style on all current and future elements as such
- As per our style guide, all panel at bottom(except model popup) should be have white background, no shadow and border on top.
- Related core PR to add wrapper on specific page, https://github.com/civicrm/civicrm-core/pull/11096


scss/civicrm/administer/civi-case/_case-types.scss
```diff
@@ -1,3 +1,13 @@
 &.page-civicrm-a {
   @include civihr_table();
+
+  table {
+    tr.forked {
```
meladawy: Can u reduce the hierarchy ? 
table tr.forked {

scss/civicrm/administer/civi-case/_case-types.scss
```diff
@@ -1,3 +1,11 @@
 &.page-civicrm-a {
   @include civihr_table();
+
+  table tr.forked {
+      font-weight: normal;
+  }
+
+  #crm-main-content-wrapper .crm-submit-buttons {
+    box-shadow: none;
```
AkA84: isn't this supposed to be something applied to *any* button in the core screens? why do you have to specify it for this particular button?

scss/civicrm/administer/civi-case/_case-types.scss
```diff
@@ -1,3 +1,11 @@
 &.page-civicrm-a {
   @include civihr_table();
+
+  table tr.forked {
+      font-weight: normal;
```
AkA84: wrong indentation

PR #49, opened 2017-09-08
HW-382: Styled Individual casetype setting page

mukeshcompucorp:
**Before:**
before

**After:**
after-2

**Note:** 
- Top buttons are now removed as per @jamienovick 's suggestion and approved by @colemanw 
- Related Core PR - https://github.com/civicrm/civicrm-core/pull/11096


colemanw:
@mukeshcompucorp if this screen could be improved by adding standard crm classes to it to prevent you from needing these extra style rules in shoreditch, please go ahead and submit a pr for that to civicrm-core.

jamienovick:
Also @mukeshcompucorp visually - Im not sure the section showing the top "save and cancel" should have the big block. In fact if @colemanw agrees I would remove those buttons completely in a separate PR. Having save and cancel at the bottom is enough?

colemanw:
I agree we don't need them twice, although it is a standard convention on most Civi forms to have buttons at the top and the bottom.

mukeshcompucorp:
@colemanw We do not need any global class for now.
@jamienovick we keep such big block for all pages for any crm buttons either on top or bottom though, let's decide if we want to remove it then either you or me can remove it from core file and submit a separate PR for that to @colemanw

scss/civicrm/administer/civi-case/_case-types.scss
```diff
@@ -1,3 +1,32 @@
...
+        margin: 0;
+
+        &:hover {
+          background: $gray-dark;
+          color: $crm-white !important;
+        }
+      }
+    }
+
+    .crm-block {
```
AkA84: I see this block padding being fixed multiple times, shouldn't we have a generic css rule that fixes it once and for all?

scss/civicrm/administer/civi-case/_case-types.scss
```diff
@@ -1,3 +1,32 @@
 &.page-civicrm-a {
   @include civihr_table();
+
+  .crmCaseType {
+
+    .crm-button {
```
AkA84: shouldn't any button be styled automatically at this point? Why do the buttons on this screen require specific styling? Is the markup different?

PR #54, opened 2017-09-13
AI-151: Fixed tags regression issue on contact and manage tags

mukeshcompucorp:
**Before 1:**
before-1

**After 1:**
after-1

**Before 2:**
before-2

**After 2:**
after-2

**Note:**
This PR also removes the legacy css due to structure change of civicrm core.


AkA84:
@mukeshcompucorp if you are fixing a regression issue, you need to leave details in the PR description about what the issue was and when was introduced, also

> This PR also removes the legacy css due to structure change of civicrm core.

This is not sufficient, i don't know what the legacy css was, not what structure changes (to what?) you are talking about

scss/civicrm/contact/pages/_tag-contacts.scss
```diff
@@ -109,64 +109,83 @@
...
       border-bottom: 1px solid $crm-dark-grayblue;
+      font-family: $font-family-base;
+      font-size: $font-size-h2;
+      font-weight: $crm-font-weight-h3;
+      padding: 17px 20px;
+      vertical-align: middle;
     }
 
-    #tagtree {
+    tbody td{
```
meladawy: leave a space before the opening bracket

scss/civicrm/contact/pages/_tag-contacts.scss
```diff
@@ -109,64 +109,83 @@
...
+        color: $gray-darker;
+        font-weight: $crm-font-weight-h3;
+        padding: 0;
+      }
+
+      .jstree-checkbox {
+        background-color: $crm-white;
+        background-image: none;
+        border: 1px solid $crm-darker-grayblue;
+        border-radius: 2px;
```
meladawy: Can u use a variable for radius ?

scss/civicrm/contact/pages/_tag-contacts.scss
```diff
@@ -109,64 +109,83 @@
...
+  }
+
+  #tagtree,
+  .tag-tree {
+    margin-left: -10px;
+
+    li {
+      margin: 10px;
+
+      div.jstree-wholerow-hovered,
```
meladawy: Do we have a CiviCRM class called jstree-wholerow-hovered? or is it generated by external library/ extension?

jamienovick:
Hi guys,
The functionality for this has changes as part of the cases workstream. I believe now tags show colours - @mukeshcompucorp not that these tags show with colours on the case screen. As such we should display the tags in the tag set as they would look when displayed. I think we have removed that functionality for some reason? This comes back to the fact that we need to look at CiviCRM and make it tidy within the shoreditch styling, not change functionality - as seems to have been the case here.

mukeshcompucorp:
@jamienovick Initially this was fixed as the common part of the tags which falls under Contact > Manage tags and hence, the structure got changed for this, it was considered as the regression issue but, it seems after working on the CiviCase project that, the current one is nice already as per the work flow and we should not touch it and keep it as it is just fix, margin paddings for that.

I will close this PR and create another one which will have just make tags at it is styled in case already.
That will reduce code as well.

PR #57, opened 2017-09-18
AI-159: Fixed overflow table issue

mukeshcompucorp:
**Before**
before

**After:**
after

Note:
This PR makes table layout adjustments instead of making overflow hidden by making table fit on container with table-layout: fixed;


meladawy:
This has a conflict. can u resolve this?

AkA84:
@mukeshcompucorp this seems a very regression-prone change, did you run backstopjs? Did you check personally on a handful of pages that you know contain table, so that we're sure we're not breaking any layout?

mukeshcompucorp:
@AkA84  yes, I ran backstopjs. This makes some some column hide data or divide in 2 line but, it does not break structure. 

Also, this is under wrapper,
.CRM_Contact_Form_Search_Basic,
.CRM_Contact_Form_Search_Advanced

which saves it from getting applied on any other screens

PR #81, opened 2017-10-07
Don't overlap civicrm-menu on Wordpress

mik-laj:
Follow: https://github.com/civicrm/civicrm-core/blob/dce0aa6f40bbfab20782d9c61e9f81e05ac5ce58/css/civicrmNavigation.css#L72

### DIFF
--- a/scss/civicrm/main-menu/_main-menu.scss
+++ b/scss/civicrm/main-menu/_main-menu.scss
@@ -1,7 +1,7 @@
 #civicrm-menu {
   font-family: $font-family-base;
   height: 32px;
-  z-index: 1000;
+  z-index: 99999;
 
   #crm-qsearch {
     background-color: $crm-white !important;


mattwire:
See PR #90.  Not sure why z-index has been added for that selector, it's causing a number of problems and is not defined in the original civicrm.css where it "does the right thing"

mik-laj:
It"s necessary for Wordpress. Wordpress has its own menu that this theme replaces. I will send you a screenshot as I am at home.

mattwire:
@mik-laj If wordpress works ok without the shoreditch theme then I think it will also work with my patch?  As it restores the original behaviour.

mik-laj:
You, are in minor error. Original theme have z-index.
http://i.imgur.com/y447ttx.png
And look at original theme source code:
https://github.com/civicrm/civicrm-core/blob/4.7.25/css/civicrmNavigation.css#L72
I've pasted this link already in another comment
https://github.com/civicrm/org.civicrm.shoreditch/pull/81#issue-263671425 (first messages in this issues)

AkA84:
@mik-laj this had been fixed by https://github.com/civicrm/org.civicrm.shoreditch/pull/270 , please check the PR for additional information

mattwire:
Thanks, tested in Drupal/Wordpress and it does seem to be fixed :-) Raised civicrm/civicrm-core#12632 for Joomla menu issue.

PR #86, opened 2017-10-17
HW-382: Fixed UI snagging and tabs

mukeshcompucorp:
Description:
This PR styles Individual case setting page keeping, global component of shoreditch in mind.

**Before:**
hw-382-before

**After:**
hw-382-after

**Technical Details:**
- Tabs were styled on two places (common, administrator) which we merged to have only one common style.
- Checkbox/radio placeholders are defined to allow individual checkbox/radio (i.e. without label) to have style applied

**Notes:**
- Checkbox/Radio is not styled globally keeping project time budget in mind but, it's coded the way it can be easily done in future and will extensive testing for that when doing.
- Table style is changed for padding globally

- [x] Backstop Tests passed


AkA84:
@mukeshcompucorp you have 13 files changes in one commit? I think the commit then could have been split in multiple smaller commit to keep the history of your work easier to follow

scss/civicrm/administer/civi-case/_case-types.scss
```diff
@@ -7,11 +7,42 @@
       padding: 20px;
     }
 
+    input[type=checkbox] {
```
AkA84: why do we have styling for CiviCase at all in shoreditch? Shouldn't all of this go into the extension css?

scss/civicrm/administer/common/_mixins.scss
```diff
@@ -16,8 +16,16 @@
         color: $gray-dark;
```
AkA84: why is the mixing called `civihr_table` if this is used to style core screens?

scss/civicrm/administer/civi-case/_case-types.scss
```diff
@@ -7,11 +7,42 @@
...
+      @extend %checkbox-styled;
```
AkA84: *never* use extend. Use mixins instead

scss/civicrm/common/_radio-checkbox.scss
```diff
@@ -1,3 +1,60 @@
+%checkbox-styled {
```
AkA84: use mixings, avoid using placeholders that force you to use `@extend`

scss/civicrm/common/_tabs.scss
```diff
@@ -1,40 +1,82 @@
 .ui-tabs {
-  background: #ffffff;
+  background: $crm-white;
   border-radius: 0;
   border: 0;
+  font-family: $font-family-base;
```
AkA84: why do we have to redefine the font family?

scss/civicrm/administer/common/_mixins.scss
mukeshcompucorp: @AkA84 This is from year ago and it seems the reason would be because initially it was only for civihr.

scss/civicrm/administer/civi-case/_case-types.scss
mukeshcompucorp: @AkA84 This are legacy pages of Civicase which does not have any styles loaded from civicase extension. Even templates of the extension are being called from civicrm-core instead of civicase exntention,(refer: modules/civicrm/ang/crmCaseType). Though, I will flag this to move under extension if possible to @colemanw

scss/civicrm/administer/civi-case/_case-types.scss
mukeshcompucorp: Ok

scss/civicrm/common/_radio-checkbox.scss
mukeshcompucorp: ok

PR #90, opened 2017-10-24
Remove conflicting z-index that breaks civicrm-menu when run with dru…

mattwire:
…pal admin-toolbar, civicrm uk.squiffle.kam etc.

When the z-index is specified we get issues with civicrm menu appearing behind other menus.  z-index is not defined in standard civicrm.css and everything "just works".  So unless there's a good reason for specifying it here, let's just remove it!


colemanw:
@jamienovick   :+1: for this PR. This bug has also been reported on https://civicrm.stackexchange.com/questions/21275/install-shoreditch-extension

jmcheung:
Experiencing the same symptoms on 0.1alpha17
No CiviCRM menu.

AkA84:
@mattwire the issue had been fixed by #270. Please see the PR for additional details, where the reason for the `z-index` change is explained as well

mattwire:
Thanks, tested in Drupal/Wordpress and it does seem to be fixed :-)  Raised https://github.com/civicrm/civicrm-core/pull/12632 for Joomla menu issue.

PR #91, opened 2017-10-26
AI-160: Fixed shadow and arrows of menu

mukeshcompucorp:
**Description:**
This PR,
- Adds shadow in menu
- Fixes arrows of menu to make it look like mockup

**Before 1:**
before

**After 1:**
after

**Before 2:**
before-2

**After 2:**
after-2

**Note:**
- Shadow is kept different intentionally to keep in lighter after discussion with Jon (This hides shadow when scroll position is at 0)
- Arrow are not implemented in CiviCRM so, CiviHR should remove the script of adding it.


js/crm-ui.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * This script is responsible for any common UI tweaks needed via script.
+ */
+
+(function ($, _) {  
+  $(document).ready(function () {
+    $('#root-menu-div .menu-item-arrow').each(function ($element) {
```
AkA84: put this in a documented function

PR #93, opened 2017-10-31
AI-155: Adjusted design of full-text search box

mihaelmladenov:
# Description:

### This PR styles:
* Full-text search page (/civicrm/contact/search/custom)

# Screenshots:

## Full-text search page (/civicrm/contact/search/custom)

**Before:**
before

**After:**
after

# Technical Details:
The CSS is targeted only at the search form block on the full-text search page.


scss/civicrm/full-text-search/_full-text-search.scss
```diff
@@ -0,0 +1,19 @@
+.CRM_Contact_Form_Search_Custom .crm-search-form-block table.form-layout-compressed td {
+  padding: 2px 0;
+  width: auto;
```
mukeshcompucorp: I think this is for td.label only to control width? if that's the case set it under https://github.com/civicrm/org.civicrm.shoreditch/pull/93/files#diff-21377b9c511c84c1f03c6dd1774b2ce3R11

scss/civicrm/full-text-search/_full-text-search.scss
```diff
@@ -0,0 +1,19 @@
...
+
+  label {
```
mukeshcompucorp: Can solution be achieved without changing this? only padding change should do.

mukeshcompucorp: or vertical align middle should do as well without changing any padding or margin

scss/civicrm/full-text-search/_full-text-search.scss
```diff
@@ -0,0 +1,19 @@
...
+
+  &.label {
+    font-weight: bold;
+    padding: 4px 20px;
+    vertical-align: middle;
+  }
+
+  & > input.crm-form-submit {
+    margin-left: 20px !important;
+    padding: $crm-btn-padding-small-vertical $crm-btn-padding-base-horizontal;
```
mukeshcompucorp: Can @extend .btn-sm work here is small padding is needed? left and right padding seems excessive otherwise.

scss/civicrm/full-text-search/_full-text-search.scss
```diff
@@ -0,0 +1,19 @@
...
+    margin: 0;
+  }
+
+  &.label {
+    font-weight: bold;
+    padding: 4px 20px;
+    vertical-align: middle;
+  }
+
+  & > input.crm-form-submit {
```
mukeshcompucorp: is & needed here?

PR #94, opened 2017-11-03
NDI-2: Styled shoreditch for safe colours

mukeshcompucorp:
**Description:**
This PR contains major changes about new color scheme as per WCAG.
See new color palette and designs at: 
https://invis.io/7QCECB68Z#/254761833_WCAG_AA_Options
https://invis.io/7QCECB68Z#/256894652_Contact_Summary
https://invis.io/7QCECB68Z#/256894653_Personal_Details

**Screenshots:**

#### Before
https://user-images.githubusercontent.com/26058635/32374680-c4fea3ac-c0c4-11e7-975d-6dc42ceccc82.png

#### After
https://user-images.githubusercontent.com/26058635/32374399-80cebb50-c0c3-11e7-8e17-a8414a2398c5.png

#### Before
before-2

#### After
after-2

#### Before
before-3

#### After
after-3

#### Before
before-4

#### After
after-4

#### Before
before-5

#### After
after-5

#### Before
before-6

#### After
after-6

**Notes:**
This work should be tested extensively before merging with develop branch

- [x] Backstop tested

PR #95, opened 2017-11-06
NDI-4: Styled event dashboard page

mukeshcompucorp:
### Description:
This PR styles 
- event dashboard page.
- membership dashboard page.

#### Before:
before

#### After:
after

#### Before:
before

#### After:
after

### Technical details:
- This PR fixes crm-icons in table and dropdown for for contact type (i.e. Individual, Household etc.) https://github.com/civicrm/org.civicrm.shoreditch/pull/95/files#diff-5badf14ccdff391c5b582a16a99e8412

### Notes:
- [x] browserstack tested
- [x] Backstop tested

Color palette changes are covered in https://github.com/civicrm/org.civicrm.shoreditch/pull/94


jamienovick:
@mukeshcompucorp do you think you can make the event summary title and the recent registrations titles look like part of the tables beneith them and use the grey header background? Examples: https://www.awesomescreenshot.com/image/2943612/3b4b33ad9570545ee2c8dbaa4b92364b

Can you do the same for contributions and other dashboard pages?

mukeshcompucorp:
@jamienovick yes, that can be done as well currently it's as per /civicrm/member?reset=1
should we make member dashboard as well like that? I can make some common CSS in that case for all such headers

jamienovick:
@jonscreat ^

jamienovick:
@jonscreat can you take it from here?

Yes it should be the same for all of the dashboards (and all tables basically) (i.e. Contributions, membership, events and pledges) and also for all of the tables on the contact record.

We should not have these floating headers/titles for tables. 

In some cases we have some descriptive text in between the title and the table, but we should still have one table and just have the descriptive text also in the header in a green box. Does that make sense?

mukeshcompucorp:
Yeah, that makes sense. I will change accordingly on all places.

JonScreat:
Just for clarity, we shouldn't ever have these floating headers per the styleguide.

If it's a page title it should be 'attached' to the top of the page and if it's a panel header then it should be attached to the body of the panel. 

See here for reference: https://projects.invisionapp.com/d/main#/console/10432973/221634721/preview

PR #97, opened 2017-11-08
NDI-8: Styling Event participant search results

mukeshcompucorp:
### Description:
This PR
- Styles search page of event participants.
- Moves advanced-search directory to search

#### Before:
before

#### After:
after

### Technical details:
Below code is having `!important` intentionally to override previous behaviour which is also done with `!important` in `_events.scss` file.

**Line 34-36**
```
.crm-form-submit {
  margin: 12.5px 0 0 !important;
}
```
**Line 49-51**
```
.crm-pager > .element-right {
  margin-top: 1px !important;
}
```
---
Commit https://github.com/civicrm/org.civicrm.shoreditch/pull/97/commits/ce536a0383a8fde6fd747d77d4ac974df11f884b is specifically to move `advanced-search` to common directory `search` so that, all search pages can be handled properly.
### Notes:
- [x] Backstop Tested
- [x] Browserstack Tested


scss/civicrm/advanced-search/pages/_results.scss
```diff
@@ -1,5 +1,6 @@
 .CRM_Contact_Form_Search_Advanced .crm-results-block,
-.CRM_Contact_Form_Search_Basic .crm-results-block {
+.CRM_Contact_Form_Search_Basic .crm-results-block,
```
AkA84: the folder containing this partial is "advanced-search", but here there are other type of search pages mentioned in the selectors. Can't we find a better placement for this?

scss/civicrm/contact/pages/_events.scss
```diff
@@ -26,6 +26,31 @@
...
+    }
+
+    .crm-form-submit {
+      margin: 12.5px 0 0 !important;
+    }
+  }
+
+  .crm-event-search-form-block,
+  .crm-results-block {
+    border-radius: $border-radius-base;
```
AkA84: but the radius is applied to a generic header no? Isn't that header styled somewhere else already?

PR #99, opened 2017-11-09
AI-154: Styled Search Builder Page

mihaelmladenov:
# Description:

### This PR styles:
* Search builder page (/civicrm/contact/search/builder)

### This PR changes the CSS code without affecting the styling for:
* Dashboard page (/civicrm)

# Screenshots:

### Before:
before

### After:
after

### Before:
dashboard-before

### After:
dashboard-after

# Technical Details:
Apart from styling the search builder page, this PR:
* introduces a new mixin called "section-title" to be used for styling section titles (like the "Include contacts where" title in the screenshots above);
* replaces the section title CSS of the dashboard getting started boxes with the new mixin.


scss/civicrm/search/pages/_search-builder.scss
```diff
@@ -0,0 +1,32 @@
+.CRM_Contact_Form_Search_Builder {
+  #map-field tr td {
```
AkA84: always nest
```scss
#map-field {
  tr {
    td {
      padding
    }
  }
}

File: scss/civicrm/search/pages/_search-builder.scss [hunk 2]
```diff
@@ -0,0 +1,32 @@
...
+      margin: 0;
+    }
+
+    &:before {
+      @include fa-icon($font-size-base, $fa-var-plus);
+      color: $brand-primary;
+    }
+  }
+
+  #map-field .crm-search-builder-add-row td.form-item {
```
AkA84: you already have the `#map-field` selector above, don't redefine it

scss/civicrm/search/pages/_search-builder.scss
```diff
@@ -0,0 +1,32 @@
...
+    padding-bottom: 15px !important;
+  }
+
+  #crm-submit-buttons {
+    padding: 25px 20px;
```
AkA84: this looks a very generic styling ("apply paddings to the submit buttons"), is this really necessary? don't we have this style already somewhere else?

mihaelmladenov:
@AkA84 please find below my comments:
1. Nesting of CSS rules has been fixed.
2. Combining CSS selectors has been fixed.
3. I have added to the description screenshots from the dashboard, where we are using the same "section-title". I have not found it yet at another place in CiviCRM, but since it's already used several times, I believe it's better to have a mixin for that.
4. I searched and did not find a general padding for these buttons elsewhere. On most of the other pages, the buttons are in a block together with the fields (inputs/selects/etc) above them and the whole wrapper has this padding. In our case, the content above spans from left to right and only the buttons need to have the padding on the left and right.
5. I have fixed the commit history.

jamienovick:
@mihaelmladenov does it also work for /civicrm/dashboard?reset=1? there was an issue with this before

mihaelmladenov:
@jamienovick Yes, I've changed the section titles in the dashboard page also.

PR #100, opened 2017-11-14
NDI-14: Styled status messages on event participants tasks pages

mihaelmladenov:
# Description:

### This PR styles:
* Event Participants Task Pages (spacing & border of status message at the top)
* Status messages throughout the whole website (padding of the status box & margins of first & last paragraphs as well as default background & border color for the message box)

Colors of status messages will be applied based on new palette:
https://projects.invisionapp.com/d/main#/console/12482552/261993810/preview

# Screenshots:

### Before:
before1

### After:
after3

### Before:
before2

### After:
after4

# Technical Details:
This PR changes the following two files:

### `scss/civicrm/common/_notifications.scss`:

* increased padding for all status boxes from 15px to 20px
* removed margin on top of first paragraph and margin on bottom of last paragraph in status message
* changed default background and border color

### `scss/civicrm/contact/pages/_events.scss`:

* removed margin on left of submit buttons
* removed border of status message

# Notes:
* BrowserStack tested
* Backstop.js tested


mukeshcompucorp:
@mihaelmladenov this message seems to be warning message are you sure that needs to be green?

scss/civicrm/common/_notifications.scss
```diff
@@ -101,7 +101,15 @@
 
 .crm-container > .status,
 .crm-container :not(.crm-footer) .status {
-  padding: 15px;
+  padding: 20px;
```
mukeshcompucorp: is this tested with other status on different pages?

scss/civicrm/contact/pages/_events.scss
```diff
@@ -1,3 +1,16 @@
+.page-civicrm-event-search {
+  [class*=CRM_Event_Form_Task_] {
+    .crm-form-block > .crm-submit-buttons > .crm-button:first-of-type {
+      margin-left: 0;
+    }
+
+    .status {
+      background-color: $crm-light-green-color;
```
mukeshcompucorp: can we use $alert-success-bg instead here? which is what we are trying to implement here. would be slightly different for now but, will be covered up sooner

scss/civicrm/contact/pages/_events.scss
```diff
@@ -1,3 +1,16 @@
...
+      border: 0;
```
mukeshcompucorp: Why we need to have border: 0; can't this be like alert component like in style guide?

mihaelmladenov:
@mukeshcompucorp, please find below my responses:

1. This is what the JIRA ticket says regarding the background color - please see https://compucorp.atlassian.net/browse/NDI-14

2. I have tested the status messages on the other pages and everywhere we have 20px padding now. Going through the Style Guide, I do not see a specific design for such in-page status messages, I can only see alerts.

3. The background color & border color of the status message have been changed.

JonScreat:
@mihaelmladenov @mukeshcompucorp We've only just introduced the yellow background for these as seen here: https://projects.invisionapp.com/d/main#/console/12482552/261993810/preview

If we could follow this styling for the colours going forward that would be great

mihaelmladenov:
@JonScreat This means we leave them yellow?

JonScreat:
@mihaelmladenov They should be yellow, yes - but the one set out in the invision link I just sent rather than how they currently are

mukeshcompucorp:
@mihaelmladenov for the padding as well you will find that in mockup

mihaelmladenov:
@mukeshcompucorp padding, background and border colors have been fixed.

scss/civicrm/common/_notifications.scss
```diff
@@ -101,7 +101,18 @@
 
 .crm-container > .status,
 .crm-container :not(.crm-footer) .status {
-  padding: 15px;
+  background-color: $alert-warning-bg;
+  border-color: $brand-warning;
+  color: $gray-darker;
+  padding: 16px 20px;
+
+  p:first-child {
```
AkA84:
```scss
p {
  &:first-child {

  }

  &:last-child {

  }
}

File: scss/civicrm/contact/pages/_events.scss [hunk 5]
```diff
@@ -1,3 +1,15 @@
+.page-civicrm-event-search {
+  [class*=CRM_Event_Form_Task_] {
+    .crm-form-block > .crm-submit-buttons > .crm-button:first-of-type {
+      margin-left: 0;
```
AkA84: where did the margin come in the first place?

scss/civicrm/contact/pages/_events.scss
```diff
@@ -1,3 +1,15 @@
...
+    }
+
+    .status {
```
AkA84: why is the border removed only on this specific page?

AkA84:
> Going through the Style Guide, I do not see a specific design for such in-page status messages, I can only see alerts.

@mihaelmladenov what style guide?

> Colors of status messages will be applied based on new palette.

Add a link to the new palette

scss/civicrm/contact/pages/_events.scss
mihaelmladenov: I have added a border to all status messages within the page, but Jon advised that for status messages that are inside the pane itself we should not have a white border. So that's why I put it here. Shall this be handled in another way?

scss/civicrm/contact/pages/_events.scss
mihaelmladenov: The left margin comes from core civicrm CSS. On other pages, such as the form for adding a new activity, the left margin has been removed in a similar way.

AkA84: then why aren't we doing this in a more generic way, instead of doing it on a page-by page basis?

scss/civicrm/contact/pages/_events.scss
AkA84: > I have added a border to all status messages within the page

Can you show me a screenshot of what status message you are talking about in this case?

> but Jon advised that for status messages that are inside the pane itself we should not have a border

then why aren't we doing this in a more generic way, instead of doing it on a page-by page basis?

scss/civicrm/contact/pages/_events.scss
mihaelmladenov: To be honest, I have absolutely no idea why that was done in this way. It's my first work on these buttons, so will try to implement a more generic solution.

jamienovick: There is a lot of this from @mukeshcompucorp first works on the theme in Feb - May of last year - perhaps he can clarify?

scss/civicrm/contact/pages/_events.scss
mihaelmladenov: I have found this at this page, but as soon as it's loaded it's removed from here and made into an alert on the top right:
message-outside-pane

I checked quite a lot of pages and could not find really an example for a status message outside the pane. Can you confirm it they are always inside the pane? If that's the case I can remove the border.

jamienovick: Guys hold on - these are the warning/notification messages that get populated as the popup in the top right of CiviCRM. Because of the fact that these were created in various different ways there is some JS that grabs them from these divs and then shows them as a popup in the right hand top corner. Lets not break these please!! Don't change existing CiviCRM functionality!!

scss/civicrm/contact/pages/_events.scss
AkA84: @mihaelmladenov any style that is supposed to be generic cannot be redefined over and over for each page. if it was done like that (and it's very likely it is the case) then the style needs to be refactored so that the style is applied only once and we don't have to go back on it for every page) @mukeshcompucorp please provide assistance if necessary

scss/civicrm/contact/pages/_events.scss
AkA84: i'm not sure i understand: we are adding a border for an element that we can't even find? @JonScreat @mukeshcompucorp can you help @mihaelmladenov here?

jamienovick: see here:
image
Thats the same element that Mihael is talking about

JonScreat: The mockup is for the element i've highlighted in blue here
image

scss/civicrm/contact/pages/_events.scss
mukeshcompucorp: @mihaelmladenov @jamienovick this component is already made generic at https://github.com/civicrm/org.civicrm.shoreditch/blob/master/scss/civicrm/common/_buttons.scss#L108

Though margin does not get applied on all the pages so, other pages styled are fine without adding margin-left: 0. but, anything related to that style should be written in this file. so, it can prevent margin on other such pges as well.

scss/civicrm/contact/pages/_events.scss
AkA84: >  @AkA84 @JonScreat the style guide extension is therefor wrong and we need to make it clear what developers should be following.

@jamienovick i don't know how it is wrong, i'm not even sure what element @mihaelmladenov is talking about . You are talking about the notification popup while @JonScreat is talking about a standard alert message

mukeshcompucorp: @jamienovick @AkA84 @mihaelmladenov 

@jamienovick you are correct. this should not get changed when we do any styling.

@mihaelmladenov the style in this ticket should be applied only to warning message which does not get converted to such popup. you will get class .no-popup when a pop is not going to be converted in popup
you can confirm that on the page you are stlying

JonScreat: The ticket talks about this item: 
image

So this is the only element I've been talking about.

My advice was on changing the colour to the alert yellow from here: https://projects.invisionapp.com/d/main#/console/12482552/261993810/preview
(see the items at the bottom of the page) 

But that we shouldn't have a border around it in this instance as it sits within the panel and that only ones that sit on the page background should have a border.
e.g. the element outlined in blue here: 
image

I've not spoken about functionality of any of the pop-ups in the corner - we have a style for these in the current styleguide, and we shouldn't be changing that. Just updating the colours of the text and icons to match the new ones.

mihaelmladenov: @jamienovick @JoeMurray @AkA84 

After a call with Alessandro, it became clear that the confusion came from miscommunication. So what I was talking about was always the yellow box on the Cancel Event Participants page. The screenshot I sent of the Manage Groups page was just an example, where we have the same wrapper element, but it gets turned into a popup. So by altering the styles for the yellow box, I am not altering at all the popups.

We agreed with Alessandro to change the background color and border of the status box to the new color palette globally and to remove the border of the status box only when it is inside a panel.

I guess this solves the confusion. Will try to be more careful in communicating from now on.

AkA84: ok @jamienovick @JonScreat turned that none of those elements are the subject of this PR here

---

@jamienovick the elements that are transformed to into a popup are elements that match this selector
```js
$('div.messages:visible').not('.help').not('.no-popup')
```
so elements that have the `.messages` class but neither the `.help` nor the `.no-popup` class

moreover, when the element is put in the notification, it gets stripped of the `.status` and `.messages` classes
```js
$(this).removeClass('status messages');
```
@mihaelmladenov here is styling the `.status` class, so the notification popup will not be affected

---

@JonScreat the green alert/message here
help
it's *not* the same as this alert/message here
status
but @mihaelmladenov did not catch that, so you were basically talking about two different elements without realizing it

The second element cannot appear directly on the background, as that's the kind of element that will be grabbed by JS and transformed into a popup notification (see above)

scss/civicrm/common/_notifications.scss
```diff
@@ -99,9 +99,31 @@
...
+      }
+
+      &:last-child {
+        margin-bottom: 0;
+      }
+    }
+  }
+
+  .crm-block,
+  form {
```
mihaelmladenov: `form` is needed for status messages that are directly inside forms, for example when deleting contributions.

mihaelmladenov:
@AkA84 The backstop.js test showed no issues at all from changing the left margin of the submit buttons. I have attached it, just in case:

[html_report.zip](https://github.com/civicrm/org.civicrm.shoreditch/files/1476937/html_report.zip)

PR #102, opened 2017-11-20
NDI-18: Fixed hidden labels on view contact screen

mukeshcompucorp:
**Description:**
This PR fixes view contact screen to make label visible

**Before 1:**
image

**After 1:**
image

**Before 2:**
image

**After 2:**
image

Tests:
- [x] Backstop test passed
- [x] Browserstack passed


AkA84:
@mukeshcompucorp do you know why they were hidden in the first place?

mukeshcompucorp:
@AkA84 no but, it was breaking if we just remove `display: none;` for labels and fields like below,
image

Also it was done as the part of PCHR-1558 at https://github.com/civicrm/org.civicrm.shoreditch/commit/6e9d5bab2f265efa7f5e2c71a08a5596a54e012c which seems purely unintentional.

PR #103, opened 2017-11-21
NDI-23: Styled campaign dashboard page

mukeshcompucorp:
## Description:
- This PR styles campaign dashboard page.

### Before:
image

### After:
image

## Technical details:
- `scss/civicrm/common/_common.scss` file adds `@import 'editable-content';` to reflect inline edit form design to CiviCRM.
- `.dataTables_wrapper {` is introduced at line `336`  of file `scss/civicrm/common/_tables.scss` to handle padding thought CiviCRM for datatables
- New directory `campaign` is introduced and added `@import "campaign/campaign";` to handle all pages and style related to campaign.

## Comments:
- [x] BackstopJS tests passed
- [x] Browserstack tested


scss/civicrm/campaign/_campaign.scss
```diff
@@ -0,0 +1,43 @@
+.page-civicrm-campaign {
+  .crm-content-block {
+    @include box-shadow($box-shadow);
```
AkA84: why just for this page?

scss/civicrm/campaign/_campaign.scss
```diff
@@ -0,0 +1,43 @@
...
+      box-shadow: none;
+    }
+
+    .dataTables_wrapper {
+      margin-bottom: 0;
+
+      > table {
+        thead th:nth-child(2),
+        tbody td:nth-child(2) {
+          padding-left: 20px;
```
AkA84: why just for this page?

scss/civicrm/campaign/_campaign.scss
```diff
@@ -0,0 +1,43 @@
...
+        }
+      }
+    }
+
+    > #mainTabContainer {
+      border-radius: $border-radius-base;
```
AkA84: why just for this page?

scss/civicrm/campaign/_campaign.scss
```diff
@@ -0,0 +1,43 @@
...
+
+      .ui-tabs-nav {
+        border-radius: $border-radius-base $border-radius-base 0 0;
+      }
+
+      .ui-tabs-panel {
```
AkA84: why just for this page?

scss/civicrm/common/_editable-content.scss
```diff
@@ -0,0 +1,6 @@
+@import '../bootstrap/components/editable-content';
```
AkA84: what is this for?

scss/civicrm/campaign/_campaign.scss
mukeshcompucorp: We generally do not give shadow to this section.
This page is exception having tab element as main and other sections like table inside that tab content. so, we are adding shadow to this page and remove it from inner elements.

scss/civicrm/campaign/_campaign.scss
mukeshcompucorp: First td should have padding-left in normal case but, in this page first td is hidden so, we need to target second one.

scss/civicrm/campaign/_campaign.scss
mukeshcompucorp: outer most element should have 2px border radius, as this page is having tab container as outer element, we give radius here

scss/civicrm/campaign/_campaign.scss
mukeshcompucorp: outer container should have 20px padding + elements like accordion or table should stick to this panel. so, we are adding this in specific to this page.

scss/civicrm/common/_editable-content.scss
mukeshcompucorp: This is inheritance of editable content from bootstrap which shares same structure of markup.
image

scss/civicrm/campaign/_campaign.scss
AkA84: can't we make it generic then, for any page with the "tab element as main"?

scss/civicrm/campaign/_campaign.scss
AkA84: can't we make it generic then, for every page with "having tab container as outer element"?

scss/civicrm/campaign/_campaign.scss
AkA84: can't we make it generic then, for any page with the structure of this specific page?

scss/civicrm/campaign/_campaign.scss
mukeshcompucorp: No actually, these elements are not having any such containers which we can target.
In this page as well see `.crm-content-block`, this is very generic class which in major case should not have any shadow.

So, I targeted it specific to this page.

scss/civicrm/campaign/_campaign.scss
mukeshcompucorp: No as I this pages are rare and there is not real identification that can tell us to recognize that tab container is outer element.

scss/civicrm/campaign/_campaign.scss
mukeshcompucorp: no this is much riskier given that we do not have any parent class identifying to have page structure is like this.

In general cases, tab panels do have content or forms, in our case we have tables and accordions.
Also, all elements are having generic classes so, it makes it more problematic to handle by creating general scenarios.

scss/civicrm/campaign/_campaign.scss
AkA84: why having a container is important? wouldn't it be enough to see, for example, if a tab element is a direct descendant of a given element?

mukeshcompucorp: Seems good idea. I'ill work out on this and will let you know.

scss/civicrm/common/_base.scss
```diff
@@ -263,3 +263,45 @@
     color: $gray-darker;
   }
 }
+
+.page-civicrm:not(.page-civicrm-contact-view) {
```
AkA84: so i can guess that this is the generic style for pages with a certain structure, but it's basically impossible to understand that from just looking at the file. I think this needs either a separate partial, or to be placed in a already existing partial that at least is pertinent with what the style is about. Either way, it definitely needs some comment with details

scss/civicrm/common/_tables.scss
```diff
@@ -333,3 +332,20 @@ table {
     @extend %civihr-table;
   }
 }
+
+.dataTables_wrapper {
```
AkA84: can you explain here what this generic style is for? so we can figure out how to document it

scss/civicrm/common/_base.scss
mukeshcompucorp: Ok I'ill do it accordingly with proper comments

scss/civicrm/common/_tables.scss
mukeshcompucorp: This is for data tables, any data table in CiviCRM should have padding as defined in here.
CiviCRM does have tables for form as well. so, we are targeting only tables which are datatable using this wrapper and giving accurate padding to it.

AkA84: do you think a `_datatables.scss` partial would make sense? what's your opinion?

mukeshcompucorp: Yeah, that makes sense. 
I'm also planning to move other style of data tables to that file when it comes to any related ticket so, we can have it on one file currently it is having it defined on different pages.
Though, I'm not doing all things in one step so, things stay more clear when we do it.

PR #113, opened 2017-12-01
NDI-5: Fix events css glitches

mukeshcompucorp:
## Description
- This PR fixes css of events page.

## Screenshot

**Before:**
screenshot-1 2
screenshot-2 2
screenshot-3 3

**After:**
after

## Technical Details
- This PR changes alpha filters style of search result globally,
```scss
#alpha-filter {
  li {
    color: $crm-gray-matte;
    font-size: $font-size-base;
    font-weight: $crm-font-weight-h3;

    a,
    a:link {
      border-bottom: none;
      color: $brand-primary;
      font-weight: $crm-font-weight-h3;

      &:hover {
        color: darken($brand-primary, $crm-darken-percentage);
      }
    }
  }
}
```
## Notes
- [x] BackstopJS tested


scss/civicrm/common/_base.scss
```diff
@@ -315,3 +315,23 @@ a.sorting_asc::before {
 a.sorting_desc::before {
   @include fa-icon($font-size-base, $fa-var-sort-desc, $brand-primary);
 }
+
+#alpha-filter {
```
AkA84: why not then a separate partial? we can't cram everything into `_base`

scss/civicrm/common/_base.scss
```diff
@@ -315,3 +315,23 @@ a.sorting_asc::before {
...
+  li {
+    color: $crm-gray-matte;
+    font-family: $font-family-base;
+    font-size: $font-size-base;
+    font-weight: $crm-font-weight-h3;
+
+    a,
+    a:link {
```
AkA84: why is `a:link` important to specify?

scss/civicrm/contact/pages/_events.scss
```diff
@@ -711,10 +711,67 @@
...
+        }
+      }
+
+      .crm-event-searchevent-form-block-title {
+        input[type=text] {
+          width: 410px;
+        }
+      }
+
+      #s2id_event_type_id {
```
AkA84: what's this `s2id` prefix? seems dynamically generated

mukeshcompucorp: Yes that's added for any select2 select dropdown. a necessity here as there is no other surrounding unique class or ID

scss/civicrm/common/_base.scss
mukeshcompucorp: That's due to core CSS file it overrides :link color over normal a tag and we need to prevent that.

scss/civicrm/common/_base.scss
mukeshcompucorp: Ok, I will fix this

PR #114, opened 2017-12-01
NDI-25: Style Survey add/edit/configure page

mukeshcompucorp:
## Description
- This PR fixes survey pages

## Screenshots
**Before:**
image
**After:**
image

**Before:**
image
**After:**
image

**Before:**
before-3
**After:**
after-3

## Technical details
Bootsrap forms variable is overridden to take placeholder color as per new color palette.
```scss
$input-color-placeholder:   $crm-gray-matte !default;
```
## Notes
- [x] BackstopJS tested


scss/civicrm/campaign/_campaign.scss
```diff
@@ -10,3 +10,5 @@
     }
   }
 }
+
+@import "pages/survey";
```
AkA84: this can't be here, we should have `@import`s only in the main file

mukeshcompucorp: We include sub parts of the module like this
i.e. `search/pages` are included via `search/_search.scss`
This helps us to have clear view on how sub-parts are managed.
see: https://github.com/civicrm/org.civicrm.shoreditch/blob/develop/scss/civicrm/search/_search.scss

PR #120, opened 2017-12-12
HW-449: Fix select box style

mukeshcompucorp:
## Description
This PR
- Styles select2 dropdown to match font size and border color

## Screenshots
**Before:**
before-1
**After:**
after-1

## Tests
- [x] BackstopJS tested


scss/bootstrap/vendor/select2/select2-bootstrap.scss
```diff
@@ -18,7 +18,7 @@
 }
 
 .form-control.select2-container.select2-dropdown-open {
-    border-color: #5897FB;
+    border-color: $brand-primary;
```
AkA84: you are modifying vendor files, you can't do it

PR #121, opened 2017-12-14
HW 451: Fixed footer style to match WCAG AA

mukeshcompucorp:
## Description
This PR,
- Styles footer based on new palette
- Fixes all white `#FFFFFF` color with variable $crm-white.

### Footer
**Before:**
image
**After:**
image

## Technical Details
All static white color `#FFF` or `#FFFFFF` is now $crm-white variable
Examples,
```diff
-  color: #fff;
+  color: $crm-white;
```
```diff
-          border-left-color: #ffffff;
+          border-left-color: $crm-white;
```
## Notes
- PRs are separated to give clear view of changes

## Tests
- BackstopJS tested
- WCAG AA standards tested for footer


scss/civicrm/common/_crm-footer.scss
```diff
@@ -0,0 +1,13 @@
...
+  a,
+  a:link {
+    color: darken($brand-primary, $crm-darken-percentage);
+  }
+  .crm-error {
+    border-radius: 2px;
+  }
+}
+
+#civicrm-footer {
```
AkA84: why are we using both `.crm-footer` and `#civicrm-footer`? What's the difference? Can't we use just one?

scss/civicrm/common/_crm-footer.scss
```diff
@@ -0,0 +1,13 @@
+.crm-footer {
+  a,
+  a:link {
+    color: darken($brand-primary, $crm-darken-percentage);
+  }
+  .crm-error {
```
AkA84: missing an empty line

scss/civicrm/common/_crm-footer.scss
```diff
@@ -0,0 +1,13 @@
...
+    border-radius: 2px;
```
AkA84: isn't the radius value already a variable?

scss/civicrm/common/_crm-footer.scss
mukeshcompucorp: This border style is done with ID in core civicrm so, we are handling it by doing it with ID rather than using !important.

AkA84: then can't you use `#civicrm-footer` for all the styling, it's confusing that we are using two selectors with similar names

mukeshcompucorp: Ok I will do that way. It can be done.

PR #126, opened 2017-12-26
NDI-34: Admin system settings styling

mukeshcompucorp:
# Description
This PR styles all pages under,
Administer > System Settings

# Screenshots
_Samples only as other pages are having similar changes_
**Before:**
image
**After:**
image

**Before:**
image
**After:**
image

**Before:**
image
**After:**
image

# Tests
- BackstopJS tested
- Browserstack tested


scss/civicrm/administer/settings/_connection.scss
```diff
@@ -0,0 +1,8 @@
+.crm-connection-block {
+  @include civicrm-table;
+
+  h3 {
+    @include panel-header;
+    margin-bottom: 0;
+  }
+}
```
AkA84: set up your IDE so that it automatically leaves an empty line on save

scss/civicrm/administer/settings/_connection.scss
```diff
@@ -0,0 +1,8 @@
+.crm-connection-block {
+  @include civicrm-table;
+
+  h3 {
+    @include panel-header;
+    margin-bottom: 0;
```
AkA84: why should the `h3` have no bottom margin only for this page? Why the exception?

scss/civicrm/administer/settings/_extensions.scss
```diff
@@ -1,7 +1,12 @@
 &.page-civicrm-admin-extensions {
+  .help {
+    margin: 20px;
```
AkA84: why does the `.help` component has this margin bottom only for this page? why the exception

scss/civicrm/administer/settings/_resource-urls.scss
```diff
@@ -9,13 +9,15 @@
...
     table.form-layout {
-      @include civicrm-panel;
-
       td:first-of-type {
         width: 30%;
       }
     }
 
+    input[type=radio] + label {
+      line-height: 135%;
```
AkA84: this looks extremely weird and specific

scss/civicrm/administer/settings/_sms-provider.scss
```diff
@@ -0,0 +1,14 @@
+&.page-civicrm-admin-sms-provider {
+  .crm-container {
+    .crm-content-block {
+      overflow: hidden;
```
AkA84: why only here? why the exception?

scss/civicrm/administer/settings/_smtp.scss
```diff
@@ -2,43 +2,29 @@
...
-    background: none;
-    box-shadow: none;
-
     table.form-layout-compressed {
-      @include civicrm-panel;
 
-      width: 100%;
+      td {
+        input[type=radio] + label {
+          line-height: 135%;
```
AkA84: again, this percentage value looks very weird

scss/civicrm/common/_replace-plain.scss
```diff
@@ -0,0 +1,7 @@
+.replace-plain {
+  @extend %custom-form-control, %open-sans-font-family;
+  height: auto !important;
+  line-height: normal !important;
+  min-height: 30px;
+  width: 300px;
+}
```
AkA84: same as before

scss/civicrm/common/_collapsible-block.scss
```diff
@@ -0,0 +1,31 @@
+.crm-collapsible {
```
AkA84: @mukeshcompucorp i want you to document in the PR, in a "Technical details" section, all this new common partials that you are adding. Document what they are for, show a before/after screenshot. Not of the whole page, but of the specific component that these partials are targeting

scss/civicrm/administer/settings/_connection.scss
mukeshcompucorp: Done. and set IDE to do this automatic

scss/civicrm/administer/settings/_extensions.scss
mukeshcompucorp: This is because this page is having .help section inside elements while other pages have it outside. so, we are giving margin from surrounding to make it look proper 
image

scss/civicrm/administer/settings/_resource-urls.scss
mukeshcompucorp: Removed based on new changes of radio/checkboxes.

scss/civicrm/administer/settings/_sms-provider.scss
mukeshcompucorp: Generally we have buttons on top and bottom on admin pages so, it takes radius of surrounding elements. In this case table is not taking border-radius so, we are forcing to hide and show rounded corners by doing overflow: hidden

scss/civicrm/administer/settings/_smtp.scss
mukeshcompucorp: Removed based on new changes of radio/checkboxes.

scss/civicrm/common/_collapsible-block.scss
mukeshcompucorp: Ok I will though, this component seems completely un-relevent to this PR. It's because I was working ahead in time of blockage at Christmas. I will remove and add it with appropriate PR with Documentation

scss/civicrm/administer/settings/_connection.scss
mukeshcompucorp: For all `h3` tags which are on top of page,
- If it's Inside the block, we make it look like panel-header
- If it's out side the block, we make it look like panel-header + make it's `margin-bottom: 0` to make it stick with it's nearest element. In general case, it's a crm-block.

So, this one falls under second scenario

Issue #130, opened 2018-01-08
CKEditor element shows in select2 format


sunilpawar:
CKEditor element shows in select2 format, and not showing all options on expand (on click). (z-index issue)

mukeshcompucorp:
@sunilpawar Can you please lead us to page where you are facing this issue?

AkA84:
Fixed by #218 , closing

Issue #74, opened 2017-09-25
enable-select2.js is too aggressive


colemanw:
Some selects were never intended to be Select2 and `enable-select2.js` breaks them. For example, `jquery.crmIconPicker.js` doesn't work properly.

I don't understand the rationale for this file. Why do we need to globally force this behavior? Making a select2 is very easy in CiviCRM, all you need to do is add the css class "crm-select2" to it (or in angular, the property `crm-ui-select`) and you're done. Why can't we just stick with that?

colemanw:
I keep finding breakages in the UI due to this script. Contact filters is broken in entityRef fields. The wysiwyg editor select looks weird on display preferences screen. I think this script ought to be removed before we release Shoreditch 1.0. @jamienovick what are your thoughts?

colemanw:
@mukeshcompucorp IMO the script needs to be removed. It is still causing UI breakages and everything should work fine without it.

mlutfy:
Fixed by #216

davialexandre:
Reopening, as #216 had to be removed

PR #131, opened 2018-01-08
#130 Skip CKEditor element from converting to select2 format

sunilpawar:
Use any civicrm form where ckeditor load -> Click on Link (pop up box appear) -> Click on Link Type -> It should show list.

> Before

Field converted to Select 2 Style. Drop down list not available (because of Z-index issue)
before

> After
Skipping the CKEditor element from converting to select 2 style.
Select option easily available without any issue.
after

### DIFF
--- a/js/enable-select2.js
+++ b/js/enable-select2.js
@@ -16,7 +16,10 @@ CRM.$(function () {
    * during DOM changes
    */
   var observer = new MutationObserver(debounce(function () {
-    CRM.$('select:visible:not(.no-select2):not(.crm-form-multiselect)')
+    /*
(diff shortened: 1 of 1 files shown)


mlutfy:
I can confirm this issue & fix.

Example form:

* Manage Event > Info and Settings
* Select text to link
* Try to link to an email.

colemanw:
I have recommended in #74 that we remove that script entirely, as this is just one of many places where it causes problems.

mlutfy:
Fixed by #216

mlutfy:
ping @davialexandre

AkA84:
Fixed by #218 , closing

PR #140, opened 2018-01-25
NDI-40: Fixed Import page steps for event participants

mukeshcompucorp:
# Description
This PR fixes import page steps for Event Participants

# Screenshots
**Before**
before-1
**After**
after-1

**Before**
before-2
**After**
after-2

**Before**
before-3
**After**
after-3

**Before**
before-4
**After**
after-4

# Technical Details
Function civi-page is added at `scss/civicrm/functions/_civi-page.scss`
```scss
@function civi-page($pageName) {
  @return unquote('.page-civicrm-#{$pageName}');
}
```
# Test
- BackstopJS tested
- Browserstack tested


scss/civicrm/contact/pages/_import-contact.scss
```diff
@@ -1,6 +1,7 @@
 .page-civicrm-import-contact,
 .page-civicrm-import-activity,
-.page-civicrm-import-custom {
+.page-civicrm-import-custom,
+.page-civicrm-event-import {
```
AkA84: now that i keep seeing this "drupal page" selectors all over the place, i was wondering if we couldn't abstract them in a mixin. Not only it would avoid repetition, but it would also make much clearer that we are talking about drupal (hence, CMS-specific) selector, which is something that a lot of devs miss unless they are really familiar with how drupal and civi interact with each other

So I was thinking about something like this

```scss
@mixin civi-drupal-page ($pages...) {
  $selectors: ();
  
  @each $page in $pages {
    $selectors: append($selectors, '.page-civi-#{$page}', comma);
  }
  
  #{join($selectors, ',')}{
    @content
  }
}
```
And this is how you would use it
```scss
@include civi-drupal-page('contact', 'activity', 'custom', 'import') {
  color: blue;
}
```
And this is the result
```scss
.page-civi-contact, .page-civi-activity, .page-civi-custom, .page-civi-import {
  color: blue;
}
```
I would suggest investigate the feasibility and implement it everywhere, i think it would be a huge improvement

mukeshcompucorp: @AkA84 
**Questions:**
- How do we handle `&.page-civicrm` notice the `&` sign in it?
- How to handle not part `.page-civicrm:not(.page-civicrm-contact-view)`

Apart from this it seems pretty clear to me for development and it's doable as well.

**Additional Question:**
- We have checked in past the only Drupal adds the class like `.page-civicrm` in Civi and other CMS like Joomla and WordPress is not doing so, I was wondering if we can have any solution that allows us to have theme rendered without having dependency on page-* classes. (I myself have not found any unique element though in WordPress which I already know too much from FE and BE part)

mukeshcompucorp: Apart from that, we can handle it as separate ticket as there are around 104 matches across **58 files** to do this

AkA84: @mukeshcompucorp what about something simpler but more malleable
```scss
// @NOTE there is no mention of the CMS, so we can decouple 
// the function invocation with the implementation logic
//
// (i.e. if in the future we switch cms, or we have a way to identify
// the page without using the CMS, nothing will change in terms of how
// the function is called and how it is invoked)
@function civi-page($pageName) {
  @return '.page-civicrm-#{$pageName}';
}
```
usage
```scss
#{civi-page('contact')},
#{civi-page('activity')},
#{civi-page('custom')},
#{civi-page('import')} {
  color: blue;
}

.page-civicrm:not(#{civi-page('contact-view')}) {
  color: red;
}

.some-parent {
  &#{civi-page('admin-schedulereminders')},
  &#{civi-page('admin-labelformats')},
  &#{civi-page('admin-pdfformats')}, {
    color: yellow; 
  }
}
```

mukeshcompucorp: This will work 👍

Issue #139, opened 2018-01-25
Font-weight 100 readability


hansrossel:
In the Chrome font-weight: 100 for Open-Sans is extremely thin and nearly unreadable.

I think for readability it should be at least 400 (regular) or even 700 because Open Sans is already a thin font on its own. Google fonts does not even offer Open-Sans with such an extremely thin font-weight.

<img width="485" alt="schermafbeelding 2018-01-25 om 09 37 55" src="https://user-images.githubusercontent.com/1085054/35378690-f869a388-01b3-11e8-9a3b-4434af0b6a7d.png">
<img width="299" alt="schermafbeelding 2018-01-25 om 09 41 41" src="https://user-images.githubusercontent.com/1085054/35378691-f882c6ec-01b3-11e8-93ff-f7762f2c5266.png">

hansrossel:
It's in /civicrm/_variables.scss
$crm-font-weight-td: 100 !default;

In /bootstrap/fonts/opensans.scss there is only 400, 600 and 700 defined as weight, so should use one of these.

AkA84:
hi @hansrossel , could you provide a screenshot of the issue? I agree that the definition itself of 100 is wrong and should be removed, but In my testing of the issue I could not replicate the problem, as the browser simply picks the lighter font variation available (400)

hansrossel:
There is a screenshot here above in my question above "family selected", it is quite unreadable because very light so you might not have seen it :)
I only started noticing recently that the font was so light in the tables so I suppose the latest update of Chrome is trying to make the font 100 and give us what we ask for.

hansrossel:
I suspect I'm getting 'Open Sans' 300 instead of 100 because I do have 'Open Sans' installed on my computer as a font, so the 300 is available to the system. People that do not have the font installed will then get the Bootstrap minimum 400 which you provide in Shoreditch. 

<img width="287" alt="schermafbeelding 2018-01-25 om 16 02 50" src="https://user-images.githubusercontent.com/1085054/35394865-232251d0-01e9-11e8-9845-b835e2279d3b.png">
Source: https://codepen.io/impressivewebs/pen/EIncg

AkA84:
@hansrossel ok I think the confusion is that, as far as i know, there is no 100 Open Sans. For the rest you are right, if you have the font locally it will show you 300, or if you don't have it locally it will show you 400 (like i see it, since i don't have the font locally)

In any case that value should be definitely removed , thanks for reporting this!

[...]

PR #101, opened 2017-11-16
NDI-15: Separated help box from panel on Export Event Participants Page

mihaelmladenov:
# Description:

### This PR styles:
* Export Event Participants page - it separates the help box from the main content of the page
* Visualizes the wizard bar globally
* Import Contacts page - adds specific styling for the wizard bar on this page
* Adjusts spacing around form buttons globally
* Fixes the name of a CSS file to match its function

# Screenshots:

## Export Event Participants page

**Before:**
before

**After:**
after

## Import Contributions page

**Before:**
33019222-18049184-ce02-11e7-81d0-9f00a9efc3df

**After:**
33019230-1fb6c29e-ce02-11e7-8350-ee6a4013c8a4

## Import Memberships page

**Before:**
33019172-e70f4952-ce01-11e7-94bc-83ae835325b4

**After:**
33025653-3cfa681e-ce17-11e7-84f8-f8ce9eaa2bf0

## Import Participants page

**Before:**
33019081-a04d0856-ce01-11e7-8af6-57541eb04e9c

**After:**
33019087-a67db3c4-ce01-11e7-8864-852445a6c2d9

### Step 1:

**Before:**
33018931-f50c0f64-ce00-11e7-8302-8dfafd5f4bae

**After:**
33025715-5abd07c6-ce17-11e7-8719-30e3b94b3191

### Step 2:

**Before:**
1344382581_import_contact_-_step_2_0_body_0_desktop

**After:**
1344382581_import_contact_-_step_2_0_body_0_desktop

### Step 3:

**Before:**
1344382581_import_contact_-_step_3_0_body_0_desktop

**After:**
1344382581_import_contact_-_step_3_0_body_0_desktop

### Step 4:

**Before:**
33018935-f9f1ca1e-ce00-11e7-9a61-a55c669b7800

**After:**
33018954-086a884c-ce01-11e7-8a02-790c14e8af8f

# Technical Details:
This PR changes the following files:

### `scss/civicrm/_variables.scss`:
* added a variable for the box-shadow that is used for elements inside a .crm-block that need a background color and shadow individually due to the fact that we have a help box or another element that should visually be outside of the white panel (i.e. we cannot apply the background color and shadow to the whole .crm-block div)

### `scss/civicrm/common/_buttons.scss`:
* removes left margin margin of form submit buttons and adds a default right margin

### `scss/civicrm/common/_common.scss`:
* includes the wizard styling file so that the wizard bar is visualized and styled

[...]

PR #145, opened 2018-01-30
NDI-30: Fixed issues related to tags and tree

mukeshcompucorp:
# Description
This PR:
- Styles context menu which is generated by jstree (CiviCRM component)
- Styles icon picker to make it look like unbroken input
- Removes unwanted variables and placeholders
- Styles tags background to match with new palette.
- Fixes button padding margin based on #101 in admin part.

# Screenshots
**Before**
image

**After**
image

# Technical Details
### Variable updates
below variable is removed from `scss/civicrm/_variables.scss` which is no longer needed as default weight is normal by default + it had buggy behaviours on pre-installed client machines. Ref https://github.com/civicrm/org.civicrm.shoreditch/issues/139
```
$crm-font-weight-td: 100 !default;
```
### Placeholder updates
Below place holders are removed from file `scss/civicrm/common/_input_file.scss` as they are not used anywhere.
```scss
%input-file {
  width: 300px;
  height: 120px;
  opacity: 0 !important;
  overflow: hidden;
  z-index: 20;
  padding: 0;
}
```

```scss
%label-input-file {
  position: relative;
  display: block;

  &:after {
    content: 'Drop file here \A or click to browse';
    white-space: pre;
    position: absolute;
    left: 112%;
    top: 0;
    display: block;
    width: 300px;
    height: 70px;
    border: 1px dashed $crm-page-topbar-bg;
    text-align: center;
    color: $gray-darker;
    padding-top: 50px;
    font-weight: 600;
  }

  &:before {
    content: "\f0ee";
    font-family: "FontAwesome";
    font-size: 23px;
    position: absolute;
    left: 111%;
    color: $crm-page-topbar-bg;
    top: 0;
    display: block;
    width: 300px;
    text-align: center;
    line-height: 70px;
    font-weight: 100;
  }
}
```

### New common styling
1. Context menu is styled at `scss/civicrm/common/_jstree.scss` generally so, if these is occurrence in future of such, we don't have to design it.
image

```scss
.jstree-contextmenu {
  &.vakata-context {
    @include box-shadow($box-shadow);
    background: $crm-white;
    border: 1px solid $crm-background;
    border-radius: $border-radius-base;
    padding: 0;

    li {
      > a {
        border-radius: 0;
        color: $gray-dark;

        .vakata-contextmenu-sep {
          display: none;
        }

        &:hover {
          background: $gray-light;
          box-shadow: none;
        }
      }

      &.vakata-context-hover {
        > a {
          box-shadow: none;
        }
      }
    }
  }
}
```

2. icon picker is styled at `scss/civicrm/common/_icon-picker-button.scss` to make it look like proper input then a broken one.
image

```scss
.crm-icon-picker-button {
  @include input-text;
  height: 19px;
  padding: 4px 10px;

  .ui-icon {
    margin-top: 0;
  }
}
```
# Tests
- BackstopJS tested


scss/civicrm/contact/pages/_tag-contacts.scss
```diff
@@ -16,8 +16,52 @@
     vertical-align: middle;
   }
 
+  .jstree-default {
```
AkA84: shouldn't this be part of the context-menu partial?

mukeshcompucorp: No,
There are 3 parts
1. .`jstree-default` which serves as tree like design (Only in tags)
image

2. `.jstree-contextmenu` which is slightly different version of tree (we can say different theme for tree)
image

3. `.jstree-contextmenu` in conjunction with `.vakata-context`
which is actually a context menu and we styled it at `scss/civicrm/common/_context-menu.scss`
image

AkA84: so we are talking about three different variations of jstree, as far as i understood. Then I think it deserves its own folder with a main file which `@import`s 3 partials for the 3 variations

mukeshcompucorp: actually, we are only modifying `.jstree-default` which is only for tags. there is not other recurrence of that. and anything else we are no touching. so, we do not have such separate directory. I just gave you variation to explain why the code `.jstree-default` is not belonged to context menu file.

We can do it separate if we are really designing it. but, that's not the case.

AkA84: @mukeshcompucorp you are modifying both `.jstree-default` and `.jstree-contextmenu` with `.vakata-context`, so I think that given both modifications are int he `.jstree-` family, I think it would make the code more organized to group them in a separate directory.

Or at the very least (I think we have folders for pages, not components, correct?) we should put them all in one single partial, where we can collect all these variations without scattering them around

mukeshcompucorp: yes we manage by page folders so, let me come up with some common partial.

PR #150, opened 2018-02-06
PCHR-3211: Hide labels of menu items with links

davialexandre:
### Problem

After adding the "Reports" link to the CiviHR admin menu, its label didn't get hidden when hovering the Quick Search box:

reports_menu_before

### Solution

The menu item labels are hidden use javascript. The code that does that would search for a text node that is a direct child of the menu item and wrap it in a span element with the `menumain-label` class. This works fine for menu items that don't have a link (which was the case of all the items in the menu until today), but it fails when the text node is wrapped in something else (an `a` tag, for example).

The `wrapMenuItemLabelInHTML()` method was updated to handle this better. Now, it assumes that the text node will always be next to the item icon. This way, it first searches for the icon, which can be either a direct child of the menu item or something nested inside another element, then gets the next sibling and, if it's a text note, it wraps with a span with the necessary class:

reports_menu_after


davialexandre:
Closing as this as opened against the wrong repo

PR #156, opened 2018-02-16
NDI-15: Fix CSS for help section on export participant page

mukeshcompucorp:
# Description
- This PR styles help section to look right in export participant page.

# Screenshots
**Before** 
image
**After**
image

# Comments
- BackstopJS tested


scss/civicrm/contact/pages/_export.scss
```diff
@@ -62,7 +62,7 @@
 
 .CRM_Export_Form_Map {
   .help {
-    margin-bottom: 0;
+    margin-bottom: 20px;
```
AkA84: shouldn't the `.help` element have the same margin with the rest of the page no matter what the page is?

mukeshcompucorp: Yes, .help should have margin-bottom: 20px; when out side form block(white shadowed area) and margin: 20px; when inside.
example,
image

or 
image

We are using the second approach in this form

AkA84: ok so then shouldn't the selector be generic for every kind of form, instead of being specific about the export form?

mukeshcompucorp: The problem doing that is,
Some of the screen makes it to look like help is outside the form block which is not true,
image

Such pages are having styling done different ways (from the days when, we were keeping core templates as it is and style all things just with css without html modifications).

So, we styled padding and fonts and other stuff for .help commonly but, for margins, we rely on the style of the page already have and apply accordingly.

PR #159, opened 2018-02-21
NDI-9: Fix event detail page

mukeshcompucorp:
# Description
This PR
- Introduces new mixin `civicrm-form-control` replacing placeholder ` %custom-form-control`
- Fixes margins and spacing for labels and inputs throughout CiviCRM all pages.
- Fixes textarea to be displayed as it is in terms of height and width from Vanilla CiviCRM
- Removes unwanted styling generated due to margins and spacing of labels and inputs
- Does Stylelint and fixes issues related to styling on modified files.

# Screenshots
**Before**
image
**After**
image

**Before**
image
**After**
image

**Before**
image
**After**
image

**Before**
image
**After**
image

# Notes
- The change in this PR involves almost all pages though, few of the screenshots are included here to give idea on what kind of changes has taken place in CiviCRM
- It is suggested to test this branch before merging it with master in order to make sure it does not have any other potential problem.

# Technical Details
New mixin `civicrm-form-control` is introduced at `scss/civicrm/mixins/_civicrm-form-control.scss` replacing placeholder `%custom-form-control`
```scss
@mixin civicrm-form-control() {
  @extend .form-control;
  @include open-sans-font-family();
  box-sizing: border-box;
  height: $crm-control-height;
  line-height: $crm-control-height;
}
```
---
## Margins and spacing of labels and inputs
**Note:** Below changes are affecting each and every page of CiviCRM and hence, should be noted and tested well.

### Labels
- Labels where previously had margins 10.5px on top and bottom(added by us) and `display: inline-block` which is removed to keep it spaced like it was in vanilla CiviCRM.
- All labels should be aligned the way it is in Vanilla CiviCRM so, we removed a fixed style `text-align: left` from css

This changes makes labels more concise and fit to the table as it was in Vanilla CiviCRM.

### Inputs
- We were adding 7.5px padding on top and bottom of each inputs (text,file,select,textare) which caused spacing problem allover civicrm, we removed that specific style and also removed all counter css needed due to that changes.

### Textarea
- We are removing fixed height and width from our styling to make sure that, textarea takes default height and width as it is in Vanilla CiviCRM.

# Comments
- BackstopJS tested
- Browserstack tested
- Stylelint tested


scss/civicrm/_variables.scss
```diff
@@ -1,4 +1,6 @@
+/* stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type, selector-max-id */
```
AkA84: please leave an empty line after this annotation (every file)

scss/civicrm/_variables.scss
```diff
@@ -1,4 +1,6 @@
...
 $contact-avatar: 90px;
+$crm-control-height: 30px;
```
AkA84: you are using this variable only in one place, there is no need to define it (https://sass-guidelin.es/#variables)

mukeshcompucorp: @AkA84 We are also using it on other places like https://github.com/civicrm/org.civicrm.shoreditch/pull/159/files#diff-717f96292f9fd95fa7164dfbced0bd66R140
https://github.com/civicrm/org.civicrm.shoreditch/pull/159/files#diff-717f96292f9fd95fa7164dfbced0bd66R287

There are around 6 instance where we are using this variable.

PR #162, opened 2018-02-23
NDI-41: Fixed report page layout

mukeshcompucorp:
# Description
This PR
- Styles report template instances to make it look better
- Adds padding to footer to make it look unified on all pages of civicrm
- Adds new mixin `input-fields-container`

# Screenshots
**Before:**
image
**After:**
image

**Before:**
image
**After:**
image

**Before:**
image
**After:**
image

# Technical Details
New mixin `input-fields-container` is introduces to handle all selectors for input field containers at one place,
```scss
@mixin input-fields-container {
  .form-layout,
  .form-layout-compressed,
  .crm-accordion-body,
  .crm-form-block,
  #task-section {
    @content;
  }
}
```
# Comments
- BackstopJS tested
- Browserstack tested
- Stylelint tested


scss/civicrm/common/_forms.scss
```diff
@@ -109,7 +109,8 @@
   .form-layout,
   .form-layout-compressed,
   .crm-accordion-body,
-  .crm-form-block {
+  .crm-form-block,
+  #task-section {
```
AkA84: what's going on with this list? the elements in this list seems to be very different from each (forms, accordions, sections), what is it that they all have in common?

mukeshcompucorp: This are the selectors under which inputs can be there in CiviCRM,
Basically we are targeting each selectors which has input fields, and in CiviCRM it's not always that the inputs are wrapped with `form` tag

mukeshcompucorp: Also, we can not directly style on `form` tag as, we are injecting with Resource URL so, no matter what, page it is, Admin or Public, our css will reflect and, it can have potential break on front parts

AkA84: ok could you then write a mixin that you then document? something like
```scss
@include form-containers { // find a more descriptive name
  // the style
}
```

and then mixin will be something like
```scss
@mixin form-containers {
  .form-layout,
  .form-layout-compressed,
  .crm-accordion-body,
  .crm-form-block,
  #task-section {
    @content;
  }
}

File: scss/civicrm/mixins/_input-fields-container.scss [hunk 2]
```diff
@@ -0,0 +1,11 @@
+/* stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type, selector-max-id */
+
+@mixin input-fields-container {
```
AkA84: you need to document the mixin, that was the whole point

scss/civicrm/common/_forms.scss
```diff
@@ -480,6 +379,88 @@
     }
   }
 
+  %responsive {
```
AkA84: why do we have these placeholders lying around? don't we have a folder for them?

scss/civicrm/common/_forms.scss
```diff
@@ -701,6 +682,22 @@
       color: $crm-white;
     }
   }
+
+  //Page specific
```
AkA84: shouldn't we have a partial just for this pages then?

mukeshcompucorp: No, we don't have any directory for placeholders and there are many placeholders scattered around different styles.

scss/civicrm/common/_forms.scss
mukeshcompucorp: No, we don't have any directory for placeholders and there are many placeholders scattered around different styles.

scss/civicrm/common/_forms.scss
mukeshcompucorp: Yes, that makes sense, I never had look on here. Caught this due to linter changing orders

jamienovick:
@mukeshcompucorp where CiviCRM is inconsistent could we fix the inconsistencies too with a PR to core?

mukeshcompucorp:
@jamienovick yes we are doing that as well, see my PRs [here](https://github.com/civicrm/civicrm-core/pulls?utf8=%E2%9C%93&q=is%3Apr+author%3Amukeshcompucorp).
Though, Core CiviCRM is already working fine, so, we only fix the thing which seems potentially wrong on shoreditch and civicrm for now.

scss/civicrm/common/_forms.scss
```diff
@@ -701,6 +680,7 @@
       color: $crm-white;
     }
   }
+
```
AkA84: unnecessary line

scss/civicrm/contact/_contact.scss
```diff
@@ -1,7 +1,13 @@
+/* stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type, selector-max-id */
+
 #{civi-page('contact-view')} {
   @import 'tabs';
   @import 'header';
   @import 'detail';
+
+  .select2-search::before {
```
AkA84: why is this needed only for this page?

mukeshcompucorp: This page has `margin-top: 8px` on `input[type='text']` so, it needs counter css for fixing search icon position and make it look proper.
https://github.com/civicrm/org.civicrm.shoreditch/blob/master/scss/civicrm/contact/_detail.scss#L779-L784

Screenshot:
image

AkA84: > This page has `margin-top: 8px` on `input[type='text']`

why?

mukeshcompucorp: All popups which open in Contact view popups are styled that way.
There are many popups such as below having such style. It is 2 year old styling.
image

I just moved this style from common/_forms.scss to this page specific now. In near future, when I work on popups, this will be removed and no longer needed.

scss/civicrm/common/_forms.scss
mukeshcompucorp: Fixed this now

PR #168, opened 2018-02-27
NDI-33: Fixed style for warning message of user sync

mukeshcompucorp:
# Description
This for fixes margin and border for sync user warning message.

# Screenshots
**Before**
image
**After**
image

# Comments
- BackstopJS tested
- Stylelint tested


scss/civicrm/administer/permissions/_access.scss
```diff
@@ -15,6 +17,16 @@
     .help {
       display: none;
     }
+
+    .status.no-popup {
```
AkA84: this looks some generic styling to me, why are we doing it for a specific page?

mukeshcompucorp: Status is generally converted to popups like below screenshot
image

The page we styled are rare when it comes inside such blocks with no other data and we just do styling for that pages exceptionally.
One of such exceptional variation we did at https://github.com/civicrm/org.civicrm.shoreditch/pull/155

PR #171, opened 2018-03-01
NDI-38: Fixed priceset and badgelayout styling

mukeshcompucorp:
# Description
This PR
- Fixes tables for Price set pages
- Fixes new badge layout page.
- Makes button position to default from right

# Screenshots
**Before Price set**
image
**After**
image

**Before Price set fields**
image
**After**
image

**Before  New/Edit Badge Layout**
image
**After**
after

# Comments
- Stylelint Tested
- Backstopjs Tested


scss/civicrm/administer/events/_badgelayout.scss
```diff
@@ -0,0 +1,23 @@
+/* stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type, selector-max-id */
+
+&#{civi-page('admin-badgelayout')} {
+  .crm-container {
+    h3 {
```
AkA84: i don't get it, isn't this supposed to happen automatically at this point?

scss/civicrm/administer/events/_badgelayout.scss
```diff
@@ -0,0 +1,23 @@
...
+      @include box-shadow($box-shadow);
+      @include panel-header;
+      margin-bottom: 0;
+    }
+
+    .crm-badge-layout-form-block {
+      @include box-shadow(0 8px 18px 0 rgba(48,40,40,0.25));
```
AkA84: shouldn't we have one value for box shadow across the whole theme?

jamienovick:
@mukeshcompucorp for price fields we have buttons on the right but CiviCRM has buttons on the left. Is there a reason for this change? @jonscreat?

Can you also ensure that @jonscreat is copied in on all shoreditch PR's

scss/civicrm/administer/events/_badgelayout.scss
mukeshcompucorp: This is slightly different version than standard box shadow, we use it on few pages. it's used when we want to combine detached section, to look like one.
I will create a variable for that as now, it's being use on more than one place

mukeshcompucorp:
@jamienovick on administration screens, we have buttons on right side for all the screens.
This was discussed between us in initial UI work of Administration part.
@JonScreat Please confirm

JonScreat:
@jamienovick @mukeshcompucorp I think that the positioning of such buttons is a holdover from CiviHR design. 

Looking through a number of shoreditch screens and vanilla CiviCRM screens I think it's probably more appropriate to have the buttons on the left. Otherwise we end up with a lot of empty space which looks out of place. 

I think if we are keeping layout close to vanilla CiviCRM then the button placement should be consistent with that also.

jamienovick:
Although I think CiviCRM does some slightly odd things, I don't think we should change things across the board without some careful thought.

I think its normal for the add new items, if shown on top of a panel, to be on the left. @JonScreat 

If its at the bottom, then I would expect the buttons for save and add new item to be on the right.

I think we should only change CiviCRM where we note a clear inconstancy from other screens.

jamienovick:
Although I think CiviCRM does some slightly odd things, I don't think we should change things across the board without some careful thought.

I think its normal for the add new items, if shown on top of a panel, to be on the left. @JonScreat 

If its at the bottom, then I would expect the buttons for save and add new item to be on the right.

I think we should only change CiviCRM where we note a clear inconstancy from other screens.

mukeshcompucorp:
Basically, all top and bottom buttons share same styling,
So, we would need to stick with either left or right. And if we follow Vanilla Layout then, it should on left either on top or bottom.

scss/civicrm/administer/events/_badgelayout.scss
AkA84: well can't we have a global rule then for both of these positions?

mukeshcompucorp: Actually no, if we have title inside block but on first first postion then we style it as second screenshot, just we won't need shadow in that case. and generally it's shared by same containers so. there is not way to identify it separately.

PR #172, opened 2018-03-02
NDI-8: Restore UI spinner styles

mukeshcompucorp:
# Description
- This PR restores the style for `.ui-spinner-input` to make it same as Vanilla CiviCRM again on search result pages.

# Screenshots
**Before**
image
**After**
image

# Technical Details
`.ui-spinner-input` is get overridden by Bootstrap and Shoreditch style though, it should be looking as it is without any change. So, We are again overriding all the styles back as per Vanilla CiviCRM

# Comments
- Stylelint tested
- BackstopJS tested


AkA84:
@mukeshcompucorp 

1. The PR is about a spinner, but in the screenshot I don't see any, there is a dropdown that is highlighted
2. Why is something in the bootstrap side of the theme affecting things on core screens?

mukeshcompucorp:
@AkA84
1. in Vanilla CiviCRM they call this control spinner, an input with arrows (see markup screenshot)
image

2. As this is `input[type='text']` it gets default style we did for all inputs which should not be for this specific input

scss/civicrm/search/pages/_results.scss
```diff
@@ -58,14 +60,26 @@
...
       margin-top: 7px;
     }
+  }
 
-    + .float-right {
-      float: none !important;
-      padding: 4px 10px 10px !important;
-      text-align: right;
-    }
+  .float-right {
```
AkA84: Can you show me what this is for?

scss/civicrm/search/pages/_results.scss
```diff
@@ -58,14 +60,26 @@
...
+    float: none !important;
+    padding: 10px 20px 20px !important;
+    text-align: right;
+  }
+
+  .ui-spinner-input {
```
AkA84: but this should be done in any page, right? Shouldn't this be in a more generic partial?

mukeshcompucorp: Yes, that makes sense, I will move

scss/civicrm/search/pages/_results.scss
mukeshcompucorp: See below screenshot
image

PR #173, opened 2018-03-02
NDI-38: Fix description color and button icon margins

mukeshcompucorp:
# Description
This PR
- Removes additional margins from buttons
- Fixes color for `.description` to unique color thoughout civicrm
- Fixes input size based on Vanilla CiviCRM for Badge layout page

# Screenshots
**Before**
image
**After**
image

**Before**
image
**After**
image

**Before New Contact page**
image
**After**
image

# Notes
Changes for `.description` and button margin are global and applies to all screens when used. (Both approved by @JonScreat)
# Comments
- Backstopjs tested
- Stylelint tested


scss/civicrm/administer/events/_badgelayout.scss
```diff
@@ -18,6 +18,11 @@
           min-width: 100px !important;
         }
       }
+
+      .six {
```
AkA84: what's this class?

scss/civicrm/common/_buttons.scss
```diff
@@ -26,6 +28,9 @@ $button-border-radius: 2px;
 }
 
 .crm-container {
+
+  // "Toolbar" buttons
+  $toolbar-top-space: 6px;
```
AkA84: what's this variable? why do we need it here?

scss/civicrm/administer/events/_badgelayout.scss
mukeshcompucorp: This class is applied by Vanilla civicrm when they want input to be `width: 6em` for us, I made it `width: 100px`

scss/civicrm/common/_buttons.scss
mukeshcompucorp: I haven't noticed this one as it was from stylelint. I will look on that

mukeshcompucorp:
@jamienovick 
> Not sure if it's in the scope of this pr but the upload fields for badge layouts and the width and height fields which dictate placement look very strangely positioned now.

I think we should stick for layout for Vanilla part. We can fix width of that inner table to make fields looking nearby though, it would be like, we are writing style for each line and I don't think it's good idea.
Still let me know and I will fix it.

jamienovick:
@jonscreat can you see my comment about filed placement and agree with Mukesh something sensible. @mukeshcompucorp  if we need to change the core HTML here to make something sensible then please do. No one is customising this page in CiviCRM so I think changing HTML is not a problem.

jamienovick:
Can we make it clear that the image upload fields are image upload fields? I know this is an improvement to core but seems sensible.

mukeshcompucorp:
@jamienovick I will discuss with @JonScreat and we will get back you on this.

Issue #160, opened 2018-02-21
Use portable resource URL for Custom CSS URL


themak1985:
This is super low on the priority list - but when you select shoreditch as the custom CSS URL - it generates an absolute URL instead of a portable one - which generates a notice in CiviCRM System Status.  I have avoided this by simply choosing other instead of shoreditch and entering [civicrm.files]/ext/org.civicrm.shoreditch-0.1-alpha15/css/custom-civicrm.css and the warning goes away.

davialexandre:
closing this as shoreditch now uses the new theming system which doesn't rely on the custom CSS URL anymore

PR #174, opened 2018-03-03
#160 Use portable resource URL for Custom CSS URL

adanielvv:
Before (I manually changed the url to remove the warning):
screenshot before

After:
screenshot after

See #160

### DIFF
--- a/shoreditch.php
+++ b/shoreditch.php
@@ -144,7 +144,10 @@ function shoreditch_civicrm_buildForm($formName) {
   }
   elseif ($formName == 'CRM_Admin_Form_Setting_Url') {
     $baseUrl = CRM_Extension_System::singleton()->getMapper()->keyToUrl('org.civicrm.shoreditch');
-    $cssUrl = CRM_Utils_File::addTrailingSlash($baseUrl, '/') . 'css/custom-civicrm.css';
+    $filesVarUrl = '[civicrm.files]/';
+    $filesUrl = Civi::paths()->getUrl($filesVarUrl, 'absolute');
+    $baseVarUrl= str_replace($filesUrl, $filesVarUrl, $baseUrl);
+    $cssUrl = CRM_Utils_File::addTrailingSlash($baseVarUrl, '/') . 'css/custom-civicrm.css';
     Civi::resources()
       ->addScriptFile('org.civicrm.shoreditch', 'js/urlSettingsForm.js')
       ->addVars('shoreditch', array('cssUrl' => $cssUrl));
--- a/shoreditch.php
+++ b/shoreditch.php
@@ -146,7 +146,7 @@ function shoreditch_civicrm_buildForm($formName) {
     $baseUrl = CRM_Extension_System::singleton()->getMapper()->keyToUrl('org.civicrm.shoreditch');
     $filesVarUrl = '[civicrm.files]/';
     $filesUrl = Civi::paths()->getUrl($filesVarUrl, 'absolute');
-    $baseVarUrl= str_replace($filesUrl, $filesVarUrl, $baseUrl);
+    $baseVarUrl = str_replace($filesUrl, $filesVarUrl, $baseUrl);
     $cssUrl = CRM_Utils_File::addTrailingSlash($baseVarUrl, '/') . 'css/custom-civicrm.css';
     Civi::resources()
       ->addScriptFile('org.civicrm.shoreditch', 'js/urlSettingsForm.js')


seamuslee001:
I'm a -1 on this at the moment, not everyone will necessarily be using the [civicrm.files]/ext as the location for Extensions. The current code makes sense as it will always resolve to the correct location no matter where the extensions are. I think what we may want is if it is in the files/ext we replace it to be that otherwise we don't alter it.

DanielvV:
@seamuslee001 
> not everyone will necessarily be using the [civicrm.files]/ext as the location for Extensions.

"ext" is not in my code proposal.

> The current code makes sense as it will always resolve to the correct location no matter where the extensions are.

My code proposal doesn't change that. :)

First, the absolute url to the extension is requested (original code):
`    $baseUrl = CRM_Extension_System::singleton()->getMapper()->keyToUrl('org.civicrm.shoreditch');`

Then, I get the absolute URL to `[civicrm.files]/` (my code).
`    $filesVarUrl = '[civicrm.files]/';`
`    $filesUrl = Civi::paths()->getUrl($filesVarUrl, 'absolute');`

Lastly, I replace the url part that links to CiviCRM files in `$baseUrl` with the variable.
`    $baseVarUrl= str_replace($filesUrl, $filesVarUrl, $baseUrl);`

> I think what we may want is if it is in the files/ext we replace it to be that otherwise we don't alter it.

If for some reason the extensions dir is not somewhere inside the `[civicrm.files]/` folder, `str_replace` won't alter the url. In every other case only part of the url is replaced with a variable that has the same value as the part replaced.

The url is saved in the database, so for existing installations the url will not be altered on upgrade.
Only the bullet in the settings page will jump (if the extensions are under `[civicrm.files]/*`):
If the user fixed the warning the bullet will jump to the left (see above).
Otherwise the bullet will jump to the right, making it very easy for the user to fix the warning.

seamuslee001:
Sorry Daniel, must have misread the code or something, i thought it was removing the extension base url and got confused with the screenshots, re-looking at the code i am a +1 on this, should we also be considering the other portable options?

DanielvV:
I think this is the only url that is saved in the database.

shoreditch.php
```diff
@@ -144,7 +144,10 @@ function shoreditch_civicrm_buildForm($formName) {
   }
   elseif ($formName == 'CRM_Admin_Form_Setting_Url') {
     $baseUrl = CRM_Extension_System::singleton()->getMapper()->keyToUrl('org.civicrm.shoreditch');
-    $cssUrl = CRM_Utils_File::addTrailingSlash($baseUrl, '/') . 'css/custom-civicrm.css';
+    $filesVarUrl = '[civicrm.files]/';
+    $filesUrl = Civi::paths()->getUrl($filesVarUrl, 'absolute');
+    $baseVarUrl = str_replace($filesUrl, $filesVarUrl, $baseUrl);
+    $cssUrl = CRM_Utils_File::addTrailingSlash($baseVarUrl, '/') . 'css/custom-civicrm.css';
```
AkA84: Maybe better reshuffle the code like this
```php
$filesUrlPlaceholder = '[civicrm.files]/';

$extUrl = CRM_Extension_System::singleton()->getMapper()->keyToUrl('org.civicrm.shoreditch');
$filesUrl = Civi::paths()->getUrl($filesUrlPlaceholder, 'absolute');
$baseVarUrl = str_replace($filesUrl, $filesUrlPlaceholder, $extUrl);
$legacyCssUrl = CRM_Utils_File::addTrailingSlash($baseVarUrl, '/') . 'css/custom-civicrm.css';

Civi::resources()
  ->addScriptFile('org.civicrm.shoreditch', 'js/urlSettingsForm.js')
  ->addVars('shoreditch', array('cssUrl' => $legacyCssUrl));
```

AkA84:
@adanielvv ping

MikeyMJCO:
What's needed at this point to move this issue along?

shoreditch.php
JoeMurray: @adanielvv this approach seems like an improvement

JoeMurray:
The error message is confusing because there is no option that can be changed to a variable to get rid of the error.

When a variable like [civicrm.files]/ext is used, then from one reasonable perspective, there should not be an error message displayed. So if there is a variable used in Extension Resource URL, then one might think suppressing the error would be reasonable.

However, the underlying issue is that when the site is migrated, eg from staging to production, the shoreditch theme will get confused since it has stored an absolute url in the db that is no longer valid. So it's reasonable for some sort of error to be displayed as there is an issue that site admins need to know about.

Am I right that this PR will not address problem that was intended to be addressed by using variables in resource urls, namely, site portability?

AkA84:
@adanielvv any update on this?

mlutfy:
I think this can be closed now that the CSS is managed by frontend/backend theme settings?

jamienovick:
Agreed, closing!

PR #179, opened 2018-03-07
NDI-37: Style accounts dialog table

mukeshcompucorp:
# Description
This PR styles table for accounts dialog in financial types

# Screenshots
**Before**
image
**After**
image


scss/civicrm/administer/financial/_financial-account.scss
```diff
@@ -15,4 +17,11 @@
       margin-bottom: 0;
     }
   }
+
+  #{civi-dialog()} {
+    .form-item {
+      @include civicrm-table;
+      margin: 0 -20px;
```
AkA84: if this is what happens with any table inside a dialog, shouldn't this be global

mukeshcompucorp: No not with all tables, some table can be form as well.

PR #181, opened 2018-03-08
NDI-34: Style system setting pages

mukeshcompucorp:
# Description
This PR styles System Setting pages of administration

# Screenshots
### Components
**Before**
image
**After**
image

### Extensions
**Before**
before
**After**
after

### Misc
**Before**
before
**After**
after

### SMTP
**Before**
before
**After**
after

# Comments
- BackstopJS tested
- Stylelint tested


scss/civicrm/administer/settings/_smtp.scss
```diff
@@ -10,6 +21,14 @@
       }
     }
 
+    .help {
+      margin: 20px;
```
AkA84: again, need to ask if these margins applied to such a generic element can't be standardized so that they're always the same everywhere

mukeshcompucorp: Nope,
When it's outside block like admin screens it should be `margin-bottom: 20px` which is already common.

When it's inside any block or modal it depends on how exactly it need to be margined. I can convert it to variable though like we did with status bars.

PR #183, opened 2018-03-08
NDI-25: Style profile preview buttons and survey spacing

mukeshcompucorp:
# Description
This PR 
- Fixes spacing in Results tab in survey
- Fixes profile preview buttons in all over CiviCRM

# Screenshots
**Before**
image
**After**
image

**Before**
image
**After**
image

# Comments
- Stylelint tested
- Backstopjs tested


scss/civicrm/common/_buttons.scss
```diff
@@ -338,13 +337,27 @@ $button-border-radius: 2px;
...
       position: relative;
       top: -1px;
     }
+
+    &:not([disabled]) {
+      cursor: pointer;
+    }
+
+    &[disabled] {
+      opacity: 0.65;
```
AkA84: is this something we can make generic and put in a variable?

mukeshcompucorp: @AkA84 The only other place where it is used is https://github.com/civicrm/org.civicrm.shoreditch/blob/master/base/scss/vendor/bootstrap/_buttons.scss#L50
which gets extended to other buttons already.

These 4 are the only buttons which does not have that so, I have added it as it is. When have more than one instance we will go for variable.

PR #188, opened 2018-03-09
NDI-41: Style report pages

mukeshcompucorp:
# Description
This PR styles minor glitches in report pages

# Screenshots
**Before**
image
**After**
image

# Comments
- Stylelint tested
- BackstopJS tested

### DIFF
--- a/scss/civicrm/contact/pages/_reports.scss
+++ b/scss/civicrm/contact/pages/_reports.scss
@@ -21,11 +21,25 @@ body[class*='page-civicrm-report-'] {
     }
 
     .crm-report-form-block {
+      .report-layout.statistics-table {
+        margin-bottom: 20px;
+
+        th + td {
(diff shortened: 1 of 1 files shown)


scss/civicrm/contact/pages/_reports.scss
```diff
@@ -21,12 +21,26 @@ body[class*='page-civicrm-report-'] {
     }
 
     .crm-report-form-block {
+      .report-layout.statistics-table {
+        margin-bottom: 20px;
+
+        th + td {
+          background-color: $gray-lighter !important;
```
AkA84: isn't this something that should be done for every table?

mukeshcompucorp: No there can be footers with this scenario which should not be touched.
image

PR #191, opened 2018-03-09
NDI-34: Fix radius and minor style glitches

mukeshcompucorp:
# Description
This PR 
- Adds overflow to make border radius visible on form pages
- Adds style for warning color globally
- Fixes margin for membership accounts modal

# Screenshots
**Before**
image
**After**
image

**Before**
image

**After**
image

**Before**
_Screenshot is zoomed to view difference, same for all overflows_
image

**After**
image

# Comments
- BackstopJS tested
- Stylelint tested


scss/civicrm/common/_base.scss
```diff
@@ -26,6 +26,10 @@ body {
     }
   }
 
+  .status-overdue {
```
AkA84: don't we have all the `.status-` related style all bundled together somewhere? This here seems a bit thrown in without  any structure in mind

mukeshcompucorp: No we just have 2 of such style 
another is
as per below screenshot

image

"Past Groups" is `h3` with `.status-removed` and other one is `td`.

Should I create a separate partial for text (options other than `h3`)?

jamienovick:
@mukeshcompucorp @jonscreat I think we discussed this before the but the big red and blue bars for groups also needs to go and should be a part of the header of the tables I think (i.e. like we have for memberships)? probably not for this ticket but just whilst I'm here.

PR #193, opened 2018-03-13
NDI-63: Remove bottom shadow from batch filter

mukeshcompucorp:
# Description
This PR removes bottom shadow of filter in accordion keeping only top shadow.

# Screenshots
**Before**
image
**After**
image

# Technical Details
- This change is exception to other style as, accordion is sitting alone as filter and there are no other alternate styles for page. (This is suggested and confirmed by @JonScreat)
- Variable for shadow is not used because, there is only one instance of this.

# Comments
- BackstopJS tested
- Stylelint tested


jamienovick:
@mukeshcompucorp should we be fixing the markup here rather then making an exception?

mukeshcompucorp:
@jamienovick markup is already correct in this one. The view is different and the accordion is used for filtering result and not searching. Otherwise it would be more sensible to have blue big accordion like other search pages.
So, we are writing css only

PR #194, opened 2018-03-13
NDI-62: Style contribution and expandable tables

mukeshcompucorp:
# Description
This PR
- Cleans unwanted table styles based on common styles created on already.
- Styles Contributions > Find Pledges search and result page

# Screenshots
**Before**
before-1
**After**
after-1

**Before**
image
**After**
image

# Technical Details
`scss/civicrm/mixins/_expandable-table.scss` is modified to handle all the expandable tables throughout CiviCRM

# Comments
- Stylelint tested
- BackstopJS tested


jamienovick:
@mukeshcompucorp I note the content in the rows and the table column titles are not well aligned. I note money/numbers are right aligned and text is left aligned.

Would there be a simple solution in the markup to fix this? i.e. moving the arrows to the other side or aligning the text always on the same side? What would you suggest here that allows us to be consistent and also can be applied to all tables? thanks

jamienovick:
A few other points for due course:
- I note that we have a different arrow to standard CiviCRM in terms of the dropdown/subtable collapse open, but our new arrow is not well aligned to the content and looks perhaps a little big and unclear what it does. @jonscreat can you look into this with @mukeshcompucorp 
- I note that our tables have a lot more height then standard CiviCRM. Whilst in principle that is ok we need to make sure that things look sensible if they wrap on 2 lines, so it may not be possible to keep this approach. @jonscreat please note.

mukeshcompucorp:
@jamienovick the solution for alignment is weird.
Basically, we need some identifier (class `.right` for example in `th` tag.), at present there is no such classes or indication that can identify this. I can create a PR for core file to do it though, it would be lot of tables with such problem and hence, the patch to PR would be too much time consuming as well.

Once the core is fixed, we just need to have style like,
```scss
th.right {
  text-align: right;
}
```

Though this as I said lot more time consuming to fix all this templates.
I would suggest that we keep it on later phase and, I will also fix the templates which I come across during other changes.

jamienovick:
correct me if I'm wrong but do we only have 5/6 search results screens?

- Contacts
- Contributions
- Pledges
- Membership
- Participants
- Grants

For events it seems that we have a sensible solution in that the arrows to change sort order are on the right and the title is on the left: /civicrm/event/manage?reset=1

Can we not make that change? Is this not reused code in any way that is the same across all templates?

mukeshcompucorp:
@jamienovick yes, there are 5/6 search result screens only but, tables are also used in Dashboards, admin list screens as well.

My point was about text alignment 
> I note the content in the rows and the table column titles are not well aligned. I note money/numbers are right aligned and text is left aligned.

This issue is not with arrows, I can make it on right and text on left easily. I was thinking of actually making text on right alignment like below.
image

but, if you think it's more sensible to move arrows, I will keep that decision on you and @JonScreat.

jamienovick:
@mukeshcompucorp - 

Hmm, we may be getting our wires a bit crossed.

I think my point is not that it is inconsistent between columns but actually that the column heading does not align to the column content. Without the vertical gridlines now it is hard to see clearly which heading relates to which content in the column.

I was thinking that this could be solved by moving the arrows to the right and keeping the title on the left (template change) and then some styling to ensure that all content in the table was left aligned.

Am I right in thinking what you've done is rather then forcing all the text to be left aligned is to instead leave it as is but to also align the title the same as the text in the column?

Sorry I may be missing something. 

I'm happy with either approach, but we will need to move the arrows I think either way to ensure that the text of the title and the contents of the column is aligned (if feasible).

thanks

mukeshcompucorp:
> Am I right in thinking what you've done is rather then forcing all the text to be left aligned is to instead leave it as is but to also align the title the same as the text in the column?

No @jamienovick I'm not forcing that. If you see Vanilla, you will see it to be same. There is no mechanism for header on right, or I would have done that as first thing.

Also, core template is something like below
```smarty
{foreach from=$columnHeaders item=header}
    <th scope="col">
        {if $header.sort}
            {assign var='key' value=$header.sort}
            {$sort->_response.$key.link}
        {else}
            {$header.name}
        {/if}
    </th>
{/foreach}
```
So, I can not really add a class or something in that to identify column as it should left or right.

One more thing, this is not a common template for table so, if we do something in one place, for the sake of consistency we will need to do it on all places were content is on right side as I already mentioned in one of my previous comments.

For the Arrows, I discussed with @JonSreat as well yesterday, it look more bad and, it seem that arrow is of the next column rather than of having of it's own.

jamienovick:
@mukeshcompucorp  just spoke to @jonscreat and he will give feedback. This we should left align all text and numbers and also see if we can move arrows if possible to right of column heading.

PR #198, opened 2018-03-16
NDI-59: Fix style glitches for event and campaign page

mukeshcompucorp:
# Description
This PR
 - Fixes style for new campaign page
 - Fixes dots for Manage event pages
 - Fixes width for textarea as common to adapt width of vanilla civicrm

# Screenshots
**Before**
after-1
**After**
after-1

**Before**
before-2
**After**
after-2

# Technical Details
Below code in `scss/civicrm/common/_forms.scss` is used to make width as per Vanilla CiviCRM.
```scss
&.big {
  width: 35em;
}

&.huge {
  width: 45em;
}
```

# Comments
- Stylelist tested
- BackstopJS tested


scss/civicrm/common/_forms.scss
```diff
@@ -113,6 +113,14 @@
       @include placeholder;
       box-sizing: border-box;
       width: unset;
+
+      &.big {
+        width: 35em;
+      }
+
+      &.huge {
```
AkA84: if these are the same widths in vanilla civicrm, why do we have to define them? Are we overriding them somewhere?

mukeshcompucorp: Yes, we extend `.from-control` from bootstrap which makes it  `width: 100%`.

jamienovick:
@mukeshcompucorp can we align the ellipses a little bit better to the menu title for the below:
https://user-images.githubusercontent.com/26058635/37517961-3999ee10-2939-11e8-9480-11068e1079d9.png

Also I think the manage my personal campaign page is a public CiviCRM page - i.e. for people who have created their fundraising page, so we should have minimal styles on this I think. It may be good to check in the mattermost dev room and get people's views on this.

Thanks

mukeshcompucorp:
@jamienovick Sorry, I'm not getting you about better title. Can you elaborate more on that?

For the public pages, we are already discussing that with @colemanw in our collaboration room, This is ticket like NDI-71 so, we will not be applying style to this page at all once, NDI-71 is done.

The idea is to not load our shoreditch css in such pages as we discussed few days back.

jamienovick:
Sorry yes bad comment - the ellipses - can we align them better to the label for the menu? i.e.: 
image
They just look a little bit low compare to the text?

mukeshcompucorp:
Ok that one is fixed on live screen already 
image

jamienovick:
Much better thanks!

PR #201, opened 2018-03-20
Fix #184: Prevent TypeError when quick search is not in the DOM.

### LINKED ISSUE #184: isQuickSearchOngoing assumes quick search exists in UI
This causes js errors, e.g. on a public event registration page where there is no quicksearch UI.

GinkgoFJG:


js/main-menu.js
```diff
@@ -63,7 +63,8 @@
    * @return {boolean}
    */
   function isQuickSearchOnGoing () {
-    return !!$('#sort_name_navigation').val().trim();
+    var searchValue = $('#sort_name_navigation').val() || '';
+    return !!searchValue.trim();
```
mukeshcompucorp: @GinkgoFJG please add empty line before this to make it look like
```js
function isQuickSearchOnGoing () {
  var searchValue = $('#sort_name_navigation').val() || '';

  return !!searchValue.trim();
}
```

mukeshcompucorp:
@GinkgoFJG Can you please add proper description to this PR based on our guideline at
https://github.com/civicrm/org.civicrm.shoreditch/blob/master/CONTRIBUTING.md#submitting-a-pr

js/main-menu.js
GinkgoFJG: All set.

PR #202, opened 2018-03-21
NDI-73: Fix general style snagging - part 1

mukeshcompucorp:
# Description
This ticket
- Fixes small inputs throughout Shoreditch to make it look like Vanilla in terms of width
- Fixes minor issues Addressed at [Snagging List](https://docs.google.com/presentation/d/1cf7rzx_s8wvVwXaIorerhVwk11qHCuK31QLPMTY2vz4/edit?usp=sharingl) from page 1 to 16.
- Cleans up styles

# Screenshots
**Before width of small fields**
image
**After**
image

**Before Missing Background in header**
image
**After**
image
# Notes
Only major screenshots are included, approximate 31 screens are changed for minor padding or margin issues.

# Comments
- BackstopJS tesed
- Stylelint tested


scss/civicrm/common/_forms.scss
```diff
@@ -463,6 +470,13 @@
         > div.select2-container {
           @extend %responsive;
           margin-right: 0 !important;
+
+          &.six,
+          &.eight,
+          &.crm-form-time {
```
AkA84: you are repeating this style 3 times, there must be a way to define it only once

scss/civicrm/contact/pages/_add-to-group.scss
```diff
@@ -1,19 +1,20 @@
...
         padding-left: 0 !important;
       }
     }
-  }
 
-  #id_existing_group {
-    > td {
-      padding: 20px;
+    .crm-submit-buttons {
+      margin-top: 20px;
```
AkA84: shouldn't the "submit buttons" block always have the same margin in every page?

mukeshcompucorp: Yes, Thogh on this page, We need to have gap between forms and buttons so, this one is due to that. see below screenshot
image

scss/civicrm/common/_forms.scss
mukeshcompucorp: We are using different selectors here. To make it common, we can create a mixin and load at here.
Should I create one?

AkA84: yes please, and make sure to document it. If you could follow http://sassdoc.com/annotations/ for the documentation it would be great

PR #219, opened 2018-04-12
NDI-75: Style API explorer

mukeshcompucorp:
# Description
This PR styles API explorer page and tabs of the page.

# Screenshots
**Before**
image

**After**
image

# Comments
- Stlyelint tested
- BackstopJS tested


scss/civicrm/api/_api.scss
```diff
@@ -0,0 +1,141 @@
+/* stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type, selector-max-id */
+
+#{civi-page('api')} {
+  .crm-container {
+    #mainTabContainer {
```
AkA84: why are these two selectors necessary, if we are already namespacing the style via `#{civi-page('api')}`

mukeshcompucorp: 2 reasons,
1. when we have `.page-civicrm-api` [Generated from `#{civi-page('api')}`] it skips .crm-containter wrapper which we add to all style we do for civicrm. So, we follow that on all the pages.
2. This page contains inline style like below,
```css
#mainTabContainer pre {
  line-height: 14px;
  font-size: 11px;
  margin: 0;
  border: 0 none;
}
#mainTabContainer ul.ui-tabs-nav {
  font-size: 1.1em;
  margin-bottom: .6em;
}
 ......
```
So, we are making it to override it properly.

scss/civicrm/api/_api.scss
```diff
@@ -0,0 +1,145 @@
+/* stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type, selector-max-id */
+
+#{civi-page('api')} {
+  .crm-container {
```
AkA84: @mukeshcompucorp  what about including this selector in `civi-page`? I thought we were going to do this since this selector appears every time we use this function?

Issue #214, opened 2018-04-03
Contact View - Contact Type Icon and Tooltip Summary not shown


tschuettler:
Default and custom icons indication the contact type are not shown. Normally this icon is used to show a contact summary as a tooltip.

dmaster:
![grafik](https://user-images.githubusercontent.com/19712240/38254827-8104638c-375a-11e8-97f8-7f9bad1e9104.png)
A custom contact type icon is shown at the same place, but the background image is loaded from the custom path.

current shoreditch:
Element is hidden in Contact View. 
![grafik](https://user-images.githubusercontent.com/19712240/38255365-f0be6212-375b-11e8-8b0f-9d158072b4af.png)

Default or custom icon are shown in search results:
![grafik](https://user-images.githubusercontent.com/19712240/38255640-a1ea8fde-375c-11e8-8e04-99a1e0157d40.png)

As per search results it probably should be changed to fontawesome style icons for the default contact types and alternativly loading the custom icon as background.

jamienovick:
@tschuettler please see: https://github.com/civicrm/org.civicrm.shoreditch/pull/222#pullrequestreview-113907613

I have some concerns about bringing this functionality back. Is there a specific requirement for it? I think the tooltip is not necessarily that useful and I think clutters the screen with the additional icon. I think this was a design decision taken a while back to remove it.

Im wondering why you had a need to bring this back?

tschuettler:
@jamienovick  We use it as a visual indication for the contact type and subtype.

I reported the tooltip because it appeared to be a regression for civicrm users without this theme. The listing of all groups may provide additional benefits to be seen without going to the contact tab. 
Should it be discussed with the civicrm core team to drop functionality or make it optional instead of hiding it in a theme?

PR #222, opened 2018-04-17
NDI-80: Fix tooltip visibility on contact header

mukeshcompucorp:
# Description
This PR makes contact type icon visible and fixes tooltip to look proper.
Reference: https://github.com/civicrm/org.civicrm.shoreditch/issues/214

# Screenshots
**Before**
**After**

Comments
- BackstopJS tesed
- Stylelint tested


jamienovick:
@AkA84 @jonscreat @mi-kon This change introduces quite a significant UI change from the design. An element was hidden a while ago to achieve the original design. Personally I think the system looks better without this icon, but I see the another PR has been made by another member of the community to bring this back.

From a CiviHR perspective I can image this is likely to be problematic as there are fields on this tooltip that we don't want to show and it may need to be removed in CiviHR codebase @mi-kon you may need to deal with that separately.

@mukeshcompucorp please note some issues with the tooltip itself:

http://nimb.ws/KAsFy3

PR #224, opened 2018-04-23
Cannot tell read from unread news items in dashboard

colemanw:
Overview
-----

In the stock CiviCRM theme, the dashboard shows a list of blog items. The unread items have bold titles. Shoreditch makes them all bold, which breaks usability. My proposed solution is this patch.

Before
----
screenshot from 2018-04-23 13 05 46

After
----
screenshot from 2018-04-23 13 05 18


colemanw:
@mukeshcompucorp can you please review this proposal and do the proper thing with unit tests and regenerating css files?

PR #226, opened 2018-04-26
NDI-80: Fix height for tooltip data

mukeshcompucorp:
# Description
This PR,
Fixes the height mismatch when there is no data on sections of contact summery tooltip.

# Screenshots
**Before**
image

**After**
image

# Comments
- BackstopJS tested
- Stylelint tested


jamienovick:
@mukeshcompucorp When you say "backstop JS tested" does that mean a Backstop JS for this was added to ensure that it does not break again? 

If not can we make sure that all future changes include backstop JS test to ensure that this fix does not break in future?

PR #256, opened 2018-07-23
Style up the buttons on a new traditional mailing interface in shored…

seamuslee001:
…itch

This styles up the next submit, delete, and save draft buttons on a traditional civicrm mailing interface when used with shoredich

before - buttons non active
shoreditch-before-non-active

After - buttons non active 
shoreditch-after-non-active

Before - buttons active
shoreditch-before-active

After buttons active
shoreditch-after-active

ping @mukeshcompucorp @AkA84


AkA84:
hi @seamuslee001 , thanks your PR!

Unfortunately this seems to be, on closer inspection, about the theme not having a good enough general and consistent approach with styling buttons, rather than about some issue specific with the mailings pages.

So the solution here wouldn't be to create ad-hoc fixes for mailings but rather improve the styling of buttons in general, which is something we're gonna try to do before the next release.

In case we realize that it won't be feasible then we could end up merging this PR to provide you a quick fix in the meantime, hope it sounds good!

seamuslee001:
hi @AkA84 thanks for looking into this agree don't want ad-hoc fixes perse. 

I should also note that i believe up until 5.4.0 there was no markup that differentiated significantly between e.g. the delete button, save button or the send a test buttons and next buttons. I have got https://github.com/civicrm/civicrm-core/pull/12013/commits/f097283f1b2391293a6f20727025f7095f54dc45 merged in 5.4 which does allow for some differentiation markup specifically looking at the save draft and delete buttons because at the moment they come out blue by default when you mouse over them however in Mosaico, the same buttons for save is a grey button (secondary) and the delete on is of course red. 

But yes i agree with your course of work if there is ways to do general fix that would be awesome. Let me know if you want me to review anything / test stuff out

AkA84:
Closing for now

PR #257, opened 2018-07-24
SHOR-10: Fix regression issues - find contacts

mukeshcompucorp:
# Description
This PR does the following changes in find contacts,
https://s.nimbusweb.me/share/1654819/qyfl4kxh2f8xqkg7f0zs
https://s.nimbusweb.me/share/1654838/156kyilsw1sfsr0tdud1

# Screenshots
**Before**
find contacts drupal demo 2

**After**
find contacts drupal demo

**Before**
find contacts drupal demo 3

**After**
find contacts drupal demo 4

# Comments
- BackstopJS tested
- Stylelint tested


jamienovick:
@mukeshcompucorp not sure the after looks quite correct in this case?

mukeshcompucorp:
@jamienovick, fixed

scss/civicrm/contact/pages/_tag-contacts.scss
```diff
@@ -245,3 +245,15 @@
     }
   }
 }
+
+#popupContainer {
```
AkA84: isn't `#popupContainer` generic? Why should it be under `_tag-contacts.scss` instead of being under a partial in `common/`? any style you write here will be applied to any popup, won't it?

PR #269, opened 2018-08-02
Update README.md

sleewok:
Add a note about the theme supported for drupal.


AkA84:
thanks you very much @sleewok , we plan to update the README to provide more detailed info about scope and requirements of the theme

PR #278, opened 2018-08-13
 C51-1: Fix UI and close search filters by default on Civicase Dashboard Page

swastikpareek:
## Overview
PR covers UI Fixes for Civacase Dashboard Pages.

**A. List of fixes on Civicase Dashboard Page**
1. Reduce the width of the select list for 'Show [X] Entries' to 80px
2. Reduce the font weight of the 'filter by case' dropdown to regular weight
3. Fix double horizontal line fix when filter dropdown was closed.
4. Box Shadow fix for filter opened state
5. Border width fix below the filters

**B. Generic Site wide fix**
1. Correctly aligning items of the footer panel.

 ***Note:** This fix is done globally as this was a problem site wide*

## Before
c51-1 - before - new

c51-1- before - double border issue

## After
c51-1 - after - new

#### Problem
After general fixes, on Show Dedupe contacts page the pager looked unaligned.

#### Fix
 Removing extra padding fixed the issue

#### Before
c51-1 - dedupe table before

#### After
c51-1 - dedupe table after


scss/civicrm/common/_base.scss
```diff
@@ -239,10 +239,12 @@ body {
     }
 
     .crm-datatable-pager-bottom {
+
```
AkA84: empty line

shoreditch.php
```diff
@@ -158,4 +158,8 @@ function shoreditch_civicrm_pageRun(&$page) {
   if ($pageName == 'CRM_Contact_Page_View_Summary') {
     CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.shoreditch', 'js/contact-summary.js');
   }
+
+  if ($pageName == 'CRM_Case_Page_DashBoard') {
+    CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.shoreditch', 'js/close-cases-filters.js');
```
AkA84: why do we need this script at all? Are we changing the default behaviour of the page?

swastikpareek: There is a requirement to keep the filter dropdown closed by default. So when user lands on the civicase dashboard all filter accordions are closed.

The idea way is to have `collapsed` class applied to accordion wrapper div. But doing that will need HTML changes and doing that will require core changes.

i.e. why I have implemented it through JS.

scss/civicrm/common/_base.scss
swastikpareek: Removing it.

swastikpareek: Done

Issue #110, opened 2017-11-28
CSS causing white space in Drupal


rekaeps:
I just updated shoreditch to alpha14. I immediately noticed there was white space added to the top of my Drupal pages, both on the front-end and back-end. I took a look at the developer console and it looks like 60px of padding is being added:

body:not(.toolbar) {
    padding-top: 60px !important;

I'm running a bootstrap theme for my Drupal front-end. Drupal version is 7.56, CiviCRM is 4.7.27.

![whitespace](https://user-images.githubusercontent.com/28766405/33329334-4d6c6e7c-d42a-11e7-923b-02a3ec8b5689.png)

AkA84:
@rekaeps thanks for spotting this, could you give me the exact conditions to replicate?

1. User's permissions 
2. Drupal theme used for admin
3. Any custom modules installed
4. Examples of a front-end and back-end page where the issue can be seen

Thanks!

rekaeps:
Hi @AkA84 ,

The whitespace is showing for both authenticated and anonymous users. We only have one role active, which is the default admin role. We have the default Seven admin theme, and we are using a premium theme called Jango (http://jango.nikadevs.com/?theme=jango) for our front-end. The purchased theme has its own custom modules that integrate with shortcodes (https://www.drupal.org/project/shortcode). These custom modules mostly just make use of the shortcodes so that content (views, blocks, etc.) can easily be inserted into fields.

I'm attaching a few different images of the whitespace, as it occurs on our development website. The issue seems to be focused to some pages and not other ones -- I've also attached images where the whitespace does not occur.

Thanks.

![back-end no whitespace](https://user-images.githubusercontent.com/28766405/33384859-26694842-d4f5-11e7-9b3f-f82513db3c10.png)
![back-end whitespace](https://user-images.githubusercontent.com/28766405/33384860-2676c21a-d4f5-11e7-9b67-02b93b2640ec.png)
![front-end no whitespace](https://user-images.githubusercontent.com/28766405/33384861-268a3f5c-d4f5-11e7-9a21-505b2b3afa4e.png)
![front-end whitespace](https://user-images.githubusercontent.com/28766405/33384862-269a1bd4-d4f5-11e7-9c3d-819580e966a8.png)

reswild:
[...]

PR #291, opened 2018-09-12
SHOR-73: Fix menu gap

AkA84:
Related to #110 , please see https://github.com/civicrm/org.civicrm.shoreditch/issues/110#issuecomment-412018813 for the technical details of the issue

### "Administration menu" module
before-admin-menu

###  "Administration menu Toolbar style" module
before-admin-menu-toolbar

### "Administration menu" module
after-admin-menu

###  "Administration menu Toolbar style" module
after-admin-menu-toolbar

## Technical details
There really wasn't any other way around the problem, other than increasing the specificity of the selectors. I had to restrain from doing a bit of refactoring given that the partial is used throughout civicrm and the potential for regression issues was too big for a ticket of a scope so narrow


scss/civicrm/common/_base.scss
```diff
@@ -1,17 +1,22 @@
 /* stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type, selector-max-id */
 
-body {
-  // Ensures that there is the same top padding regardless whether
-  // the user can see the admin toolbar or not
-
-  &:not(.toolbar) {
-    padding-top: $crm-main-menu-height !important;
-  }
+#{$outside-namespace-marker} {
```
mukeshcompucorp: is this variable declared somewhere? I don't see it in current PR and seems newly introduces.

AkA84: yes, it's declared here https://github.com/civicrm/org.civicrm.shoreditch/blob/master/scss/bootstrap/overrides/_variables.scss#L1

PR #292, opened 2018-09-12
SHOR-17: Fixed spacing and alignment issues

mukeshcompucorp:
# Description
This PR does the following changes in contributions,
https://s.nimbusweb.me/share/1658672/7fbgty60du9lwq57ule0
https://s.nimbusweb.me/share/1658687/bgkq6wcfp4vzffa22nmi
https://s.nimbusweb.me/share/1658702/tqkd8ad6csyf8d0tn1sv

# Screenshots
**Before**
gotv voter tracking training 2

**After**
gotv voter tracking training 1

**Before**
civigrant dashboard training

**After**
civigrant dashboard training 1

**Before**
civigrant dashboard training 2

**After**
civigrant dashboard training 4

# Note
This PR needs to have code PR merged to make proper structure [link of core](https://github.com/civicrm/civicrm-core/pull/12812)

# Comments
- BackstopJS tested
- Stylelint tested


scss/civicrm/grant/_grant.scss
```diff
@@ -48,6 +48,10 @@
         td {
           font-size: $font-size-base !important;
           padding: 15px 20px !important;
+
+          &:last-child {
```
deb1990: Cant we use a class/attribute/something else to identify the `td`? 
What if there is another td added. this will break in the case.

What is the reason of the field being right aligned in the first place.

deb1990:
@mukeshcompucorp in https://user-images.githubusercontent.com/26058635/45422119-ebb5f200-b6ac-11e8-9fce-13c8f6420115.png, the spacing between the dropdown and "Show"/"entries" look very small. Can we confirm with Jon?

scss/civicrm/grant/_grant.scss
mukeshcompucorp: We have common tables everywhere having last column to be on right aligned always but, in some cases we make it left. so, it's the right one. 

Also, the structure is core so, we don't change it until breaking or wrong something in there. we just fix with css.

mukeshcompucorp:
@deb1990 again it's just styling we do over core not the structuring like we do with CiviHR

PR #304, opened 2018-09-14
SHOR-16: Fixed membership issues

mukeshcompucorp:
# Description
This PR does the following changes for membership,
https://s.nimbusweb.me/share/1658636/rp6ka8qg1dq3cys790q4
https://s.nimbusweb.me/share/1658592/f7egeigu2x8h2hhc1r9p

# Screenshots
**Before**
find memberships training 1

**After**
find memberships training

**Before**
find memberships training 2

**After**
civimember training 13

# Note
This PR needs to have code PR merged to make proper structure [link of core](https://github.com/civicrm/civicrm-core/pull/12822)

# Comments
- BackstopJS tested
- Stylelint tested


scss/civicrm/contact/pages/_memberships.scss
```diff
@@ -423,6 +423,12 @@
       width: 105%;
     }
 
+    &.no-border {
+      margin-left: -20px;
+      margin-right: -20px;
+      width: 105%;
```
reneolivo: 105%? would this break if the screen shrinks or expands? Shouldn't it be something like `width: calc(100% - 40px);` ?

scss/civicrm/contact/pages/_memberships.scss
```diff
@@ -522,17 +532,37 @@
         }
       }
 
-      tr td.label {
-        background: none !important;
-        padding: 10px 30px !important;
+      tr {
+        th {
+          padding: 10px 20px !important;
```
reneolivo: can't we avoid adding `!important` ? is there no way to prioritize these rules without using `!important` ?

scss/civicrm/contact/pages/_memberships.scss
mukeshcompucorp: Nope: Other classes in pages are fixed like this. so, it's not breaking. see few lines above for the same.

scss/civicrm/contact/pages/_memberships.scss
mukeshcompucorp: Nope! it is not working without it, due to many already previous !important s.

mukeshcompucorp: Fixed

PR #307, opened 2018-09-14
SHOR-15: Fixed border issue

mukeshcompucorp:
# Description
This PR does the following changes for contributions,
https://s.nimbusweb.me/share/1658194/3f30wpaom35qb5xok7po

# Screenshots
**Before**
accounting batches training 3

**After**
accounting batches training 2

# Comments
- BackstopJS tested
- Stylelint tested


jamienovick:
Could the bullet point be left aligned to the text and perhaps have some consistent sizing? many thanks

jamienovick:
@reneolivo  if you are reviewing please make sure the end result looks good!! 😉 UX is bottom up!

reneolivo:
You are right @jamienovick, I'll pay more attention.

@mukeshcompucorp I think what actually needs to happen is that the container needs its padding removed.

PR #312, opened 2018-09-18
SHOR-80: Fixed line height issue

mukeshcompucorp:
# Description
This PR does the following changes for line height

# Screenshots
**Before**
new pledge training 1

**After**
new pledge training

# Comments
- BackstopJS tested
- Stylelint tested


christianwach:
The new line-heights look incorrect to me in the "after" screenshot:

line-height

jamienovick:
@deb1990 can we make sure when we review that we are checking the UX as well as the code - Should @jonscreat be reviewing these before they are merged?

mukeshcompucorp:
@christianwach this needs to be fixed in core patch and not shoreditch as it's the custom br tags added in vanilla CRM. I'm creating a patch and will update to core team about that.
See below screenshot for your reference,
image

christianwach:
Thanks for clarifying @mukeshcompucorp - perhaps this could be noted in the PR comments section to help those of us tracking Shoreditch development?

PR #315, opened 2018-09-20
SHOR-93: Exclude dist files from PRs + gulp release task

AkA84:
## Problem
A major problem with PRs in this project is that they currently must include min files (.css) in their commits (or at least their last commit) alongside the changes to the source files.

This leads to a very cumbersome situation for contributors, as the min files are a constant source of merge conflicts that the owner of a PR have to manually solve whenever some other PR gets merged in the meantime

On the other hand, keeping the min files in the repo allows to keep things simple when it comes to using the extension on production sites: the files are already included in the repo and up to date, so there is no need to have some kind of automation that builds them from source code when the extension is enabled or updated.

## Solution
This PR contains a hooks-based solution that enforces a workflow that solves both the aforementioned problem but also allows to keep the min files in the repo:

* The .css files will not be included in commits anymore. A pre-commit hook will prevent any contributor from accidentally including them by simply rejecting the commit if that happens.
* Thanks to this "forced exclusion", any given PR will then contain only changes to the source code, meaning that it will be impossible to have annoying merge conflicts on min files anymore.
* The min files will still be part of the repo, but they will be updated only right before a new release by the person in charge of the current release. A new gulp task (`gulp release`) is included in this PR to automate the process. 

    This means that any commit in-between releases will contain the css files of the previous release. This allows projects that have Shoreditch as a dependency to safely pull latest on `master` at any given point if they so desire, with the assurance that even if the source files contain untested changes, the css files are instead "stable".
* For developers and projects that instead need to use the bleeding edge version of the theme
    * running `npm i` will automatically trigger `gulp sass`
    * a post-merge git hook will do the same after any `git merge` or `git pull` that merge commits containing changes to the source files

### hooks
The hooks are simple bash scripts contained in a `.githook` folder and are automatically installed and executed by [husky](https://github.com/typicode/husky) when running `npm i`
```bash
"scripts": {
  "postmerge": "./.githooks/post-merge.sh",
  "precommit": "./.githooks/pre-commit.sh"
}
```
### gulp `release` task
```bash
gulp release --ver 0.1-alpha30
```
The above does the following:
1. Creates a `v0.1-alpha30-rc` branch
2. Runs `gulp sass`
3. Updates the _info.xml_ file with the new version and the new release date (current day)
4. Commits the min and info files, with "Update css and info files for v0.1-alpha30" as the title of the commit. The commit is created with the `--no-verify` flag so that the min files can avoid being rejected by the pre-commit hook
5. Pushes the new branch upstream
6. Shows a "success" message with the GitHub compare link

All together it looks like this:
```bash
$ gulp release --ver 0.1-alpha30
[11:23:04] Using gulpfile /path/to/gulpfile.js
[11:23:04] Starting 'release'...
[11:23:04] Starting 'release:params-check'...
[11:23:04] Finished 'release:params-check' after 591 μs
[11:23:04] Starting 'release:create-branch'...
[11:23:04] Finished 'release:create-branch' after 29 ms
[11:23:04] Starting 'sass'...
# sass sub-tasks...
[11:23:18] Finished 'sass' after 13 s
[11:23:18] Starting 'release:update-info'...
[11:23:18] Finished 'release:update-info' after 5.48 ms
[11:23:18] Starting 'release:commit-changes'...
[11:23:18] Finished 'release:commit-changes' after 65 ms
[11:23:18] Starting 'release:push'...
[11:23:22] Finished 'release:push' after 3.75 s
[11:23:22] Starting 'releasePrepared'...
Branch v0.1-alpha30-rc prepared and pushed upstream
Please check https://github.com/civicrm/org.civicrm.shoreditch/compare/v0.1-alpha30-rc
[11:23:22] Finished 'releasePrepared' after 286 μs
[11:23:22] Finished 'release' after 17 s
```

### Removed dependency on `civicrm-scssroot`
[`civicrm-scssroot`](https://github.com/totten/civicrm-scssroot) is a tool that allows cross-extension references in SASS imports, which allows `@import`s like
```scss
@import 'SCSSROOT/org.civicrm.somextensionname/scss/ALL';
```
Because the package uses [`cv`](https://github.com/civicrm/cv), it requires the extension to be placed inside a civicrm site and the dev to always have a web server running, which can become problematic if you want to run `gulp sass` on a vagrant setup. Quoting the feedback I got on the matter:

> It is really difficult to use the git hooks in a vagrant setup. Since shoreditch uses civicrm-scssroot, I am now forced to run the git commands inside my VM rather than my host machine. This means I have to setup git authentication on VM (in my case, since I use SSH keys, it means I need to copy the keys from my local system to the VM). 

Given that shoreditch doesn't reference partials in external extensions, the `civicrm-scssroot` package was removed, and the `@import`s that were using it are now leveraging the `includePaths` property of the `gulp-sass` and `gulp-sass-glob` plugins

```js
var includePaths = [
  __dirname,
  path.join(__dirname, 'scss')
];

gulp.task('sass:bootstrap', function () {
  return gulp.src('scss/bootstrap/bootstrap.scss')
    .pipe(bulk({
      includePaths: includePaths
    }))
    .pipe(sass({
      // ...
      includePaths: includePaths,
    })
```
```diff
- @import 'SCSSROOT/org.civicrm.shoreditch/scss/angular/vendor/ALL';
- @import 'SCSSROOT/org.civicrm.shoreditch/base/scss/bootstrap';
- @import 'SCSSROOT/org.civicrm.shoreditch/scss/angular/overrides/ALL';
+ @import 'angular/vendor/**/*';
+ @import 'base/scss/bootstrap';
+ @import 'angular/overrides/**/*';
````

⚠️ To maintain clarity in the `@import`s in the root files (_bootstrap.css_ and _custom-civicrm.css_), every path of any partial imported starts with the root folder of that partial.

So for example, in _bootstrap.scss`, instead of
```scss
@import 'base/scss/bootstrap';
@import 'angular/vendor/**/*';
@import 'overrides/variables';
```
It's now
```scss
@import 'base/scss/bootstrap';
@import 'angular/vendor/**/*';
@import 'bootstrap/overrides/variables';
```

### Improved `gulp watch`
Before this PR, `gulp watch` triggered the generic `sass` task for changes on any `.scss` partial
```js
gulp.task('watch', function () {
  gulp.watch(civicrmScssRoot.getWatchList(), gulp.parallel('sass'));
});
```
This means that if `gulp sass:civicrm` would be triggered even if a change occurred in a partial under the `bootstrap/` folder (which is built by `sass:bootstrap` instead)

With the removal of `civicrm-scssroot` (see above), the watch task was improved so that it triggers the correct `sass` subtask depending on which partial got updated
```js
gulp.task('watch', () => {
  gulp.watch([
    'base/scss/**/*.scss',
    'scss/angular/**/*.scss',
    'scss/jquery/**/*.scss'
  ], gulp.parallel('sass'));
  gulp.watch('scss/bootstrap/**/*.scss', gulp.parallel('sass:bootstrap'));
  gulp.watch('scss/civicrm/**/*.scss', gulp.parallel('sass:civicrm'));
});
```


.githooks/post-merge.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+YELLOW='\033[0;33m'
+NO_COLOR='\033[0m'
```
igorpavlov: Would `RESET_COLOR` be a better name here?

gulpfile.js
```diff
@@ -1,88 +1,183 @@
...
-gulp.task('watch', function () {
-  gulp.watch(civicrmScssRoot.getWatchList(), gulp.parallel('sass'));
-});
+const gulp = require('gulp');
+const argv = require('yargs').argv;
+const PluginError = require('plugin-error');
 
-gulp.task('default', gulp.parallel('sass'));
+// sass
+{
```
igorpavlov: Why do you need a separate scope here?

gulpfile.js
```diff
@@ -1,88 +1,183 @@
...
-
-  if (regex.test(selector)) {
-    selector = selector.replace(regex, function (match) {
-      return match + bootstrapNamespace;
-    }) + ',\n' + selector.replace(regex, bootstrapNamespace);
+  /**
+   * Apply the namespace on html and body elements
+   *
+   * @param  {string} selector the current selector to be transformed
+   * @return string
```
igorpavlov:
```js
@param  {String}
@return {String}
```

gulpfile.js
```diff
@@ -1,88 +1,183 @@
...
+      done();
+    }
+  ));
+
+  /**
+   * Build the YYYY-MM-DD string of today's date
+   *
+   * @return {String}
+   */
+  function formatTodayDate () {
```
igorpavlov: Any reason we do not use *MomentJS* here?

gulpfile.js
AkA84: same reason you would use a IIFE, to not pollute the outer scope with vars and functions that is not necessary to have in there

gulpfile.js
AkA84: I guess http://tonsky.me/blog/disenchantment/ hit home to me, I just didn't want to add a package as complex as momentjs just for a one-off so trivial

PR #316, opened 2018-09-20
Replace key callbacks with Symfony listeners

christianwach:
This PR addresses a limitation in the way CiviCRM extensions are built by replacing the key functional callbacks with Symfony event listeners, which have been available since CiviCRM 4.7.19. What this change makes possible is for modules, plugins and extensions to manipulate when (and if) the Shoreditch resources are loaded - something that is currently impossible.

Here's some working code which prevents Shoreditch from loading on the front-end of a WordPress site - a common request because Shoreditch's Bootstrap implementation does not play well with WordPress themes that are Bootstrap-based:

```php
class Restrict_Shoreditch_On_Front_End {

	/**
	 * Register hooks.
	 */
	public function register_hooks() {
		add_action( 'civicrm_config', array( $this, 'alter_config' ), 10, 1 );
	}

	/**
	 * Filter config when Shoreditch is active.
	 *
	 * @param array $config The config settings.
	 */
	public function alter_config( &$config ) {

		// listen for Shoreditch config complete event
		Civi::service('dispatcher')->addListener(
			'hook_shoreditch_civicrm_config',
			array( $this, 'shoreditch_config_complete' ),
			-100 // higher priority
		);

	}

	/**
	 * Remove Shoreditch callbacks from front-end.
	 *
	 * @param \Civi\Core\Event\GenericHookEvent $event The event object.
	 * @param str $hook The name of the hook currently being run.
	 */
	public function shoreditch_config_complete( $event, $hook ) {

		// bail if admin
		if ( is_admin() ) return;

		// remove Shoreditch callbacks
		Civi::service('dispatcher')->removeListener(
			'hook_civicrm_coreResourceList',
			'shoreditch_symfony_civicrm_coreResourceList'
		);

		Civi::service('dispatcher')->removeListener(
			'hook_civicrm_buildForm',
			'shoreditch_symfony_civicrm_buildForm'
		);

		Civi::service('dispatcher')->removeListener(
			'hook_civicrm_pageRun',
			'shoreditch_symfony_civicrm_pageRun'
		);

	}
	
} // class ends

// let's go
$my_test_plugin = new Restrict_Shoreditch_On_Front_End();
$my_test_plugin->register_hooks();
```

In my opinion, the nice thing about this architectural change is that it does not require maintaining code that is conditional upon the host CMS. And perhaps it even suggests a path towards a "more-OOP-less-procedural" architecture for CiviCRM extensions :-)

### DIFF
--- a/shoreditch.php
+++ b/shoreditch.php
@@ -1,5 +1,7 @@
 <?php
 
+use \Civi\Core\Event\GenericHookEvent;
+
 require_once 'shoreditch.civix.php';
 
 /**
@@ -9,6 +11,22 @@ require_once 'shoreditch.civix.php';
  */
 function shoreditch_civicrm_config(&$config) {
   _shoreditch_civix_civicrm_config($config);
+
+  // Add listeners for CiviCRM hooks that might need altering by other scripts
+  Civi::service('dispatcher')->addListener('hook_civicrm_coreResourceList', 'shoreditch_symfony_civicrm_coreResourceList', -100);
+  Civi::service('dispatcher')->addListener('hook_civicrm_buildForm', 'shoreditch_symfony_civicrm_buildForm', -100);
+  Civi::service('dispatcher')->addListener('hook_civicrm_pageRun', 'shoreditch_symfony_civicrm_pageRun', -100);
+
+  /**
+   * Dispatch an event to say that Shoreditch is configured.
+   *
+   * @since 0.1-alpha25
+   *
+   * @param string $hook_name The dispatched hook name.
+   * @param object $hook_event The dispatched hook event object.
+   */
+  Civi::service('dispatcher')->dispatch('hook_shoreditch_civicrm_config', GenericHookEvent::create(array()));
+
 }
 
 /**
@@ -114,9 +132,12 @@ function shoreditch_civicrm_alterSettingsFolders(&$metaDataFolders = NULL) {
 }
 
 /**
- * Implements hook_civicrm_coreResourceList().
+ * Implements hook_civicrm_coreResourceList() via Symfony hook system.
  */
-function shoreditch_civicrm_coreResourceList(&$items, $region) {
+function shoreditch_symfony_civicrm_coreResourceList( $event, $hook ) {
+  // Extract args for this hook
+  list( $items, $region ) = $event->getHookValues();
+
   if ($region == 'html-header') {
     CRM_Core_Resources::singleton()->addStyleFile('org.civicrm.shoreditch', 'css/bootstrap.css', -50, 'html-header');
     CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.shoreditch', 'js/crm-ui.js');
@@ -134,9 +155,12 @@ function shoreditch_civicrm_coreResourceList(&$items, $region) {
 }
 
 /**
- * Implements hook_civicrm_buildForm().
+ * Implements hook_civicrm_buildForm() via Symfony hook system.
  */
-function shoreditch_civicrm_buildForm($formName) {
+function shoreditch_symfony_civicrm_buildForm( $event, $hook ) {
+  // Extract args for this hook
+  list( $formName ) = $event->getHookValues();
+
   if ($formName == 'CRM_Contact_Form_Search_Advanced') {
     CRM_Core_Resources::singleton()->addScriptFile('org.civicrm.shoreditch', 'js/highlight-table-rows.js');
   }
@@ -150,9 +174,12 @@ function shoreditch_civicrm_buildForm($formName) {
 }
 
 /**
- * Implements hook_civicrm_pageRun().
+ * Implements hook_civicrm_pageRun() via Symfony hook system.
  */
-function shoreditch_civicrm_pageRun(&$page) {
+function shoreditch_symfony_civicrm_pageRun( $event, $hook ) {
+  // Extract args for this hook
+  list( $page ) = $event->getHookValues();
+
   $pageName = $page->getVar('_name');
 
   if ($pageName == 'CRM_Contact_Page_View_Summary') {


MikeyMJCO:
I like this - support on 4.7.19 for Shoreditch is "assumed" because Shoreditch declares `<compatibility><ver>4.7</ver></compatibility>` but in reality it's extremely unlikely that Shoreditch is properly compatible with CiviCRM <=5.0.0 and we should probably be encouraging updating to 5.3.1+ for security reasons in as many ways as possible (including through extensions).

MikeyMJCO:
Should the comments be changed to `Implements hook_symfony_civicrm_coreResourceList()` seems clearer and more accurately describes the implementation.

christianwach:
> Should the comments be changed to Implements `hook_symfony_civicrm_coreResourceList()` seems clearer and more accurately describes the implementation.

Personally I don't think so - there is no such hook, after all. The callback is simply being shifted from the `invokeViaUF()` method to a `CiviEventDispatcher` listener.

MikeyMJCO:
> Personally I don't think so - there is no such hook, after all. The callback is simply being shifted from the invokeViaUF() method to a CiviEventDispatcher listener.

Forgive my naivete here but is it necessary to change the function names to achieve this? If so then the current docs comments are okay.

But https://docs.civicrm.org/dev/en/latest/hooks/setup/symfony/ seems to indicate that it's not necessary because the event listener can be added within the hook call which prevents "breaking" the hook pattern which is `EXTENSIONSHORTNAME_HOOK()`.

christianwach:
> is it necessary to change the function names to achieve this

Yes, because otherwise those functions would *still receive callbacks* via `invokeViaUF()`. This is precisely the limitation of the `EXTENSIONSHORTNAME_HOOK` architecture that this PR addresses: namely it is impossible to unhook functions which implement that naming schema. 

What's extra sweet about the `addListener()` method is that the callback can be named whatever you want it to be named. See my sample code for an example of this.

MikeyMJCO:
In that case should we avoid the "hook_xxx" syntax altogether to avoid the confusion given that we're actually NOT implementing "hook_xxx" but rather calling a Symfony event dispatcher - we're no longer handing off to the hook implementation in core right?

christianwach:
FWIW, I'm open to suggestions on how name the callbacks. The only condition is that they *must not* be of the form `EXTENSIONSHORTNAME_HOOK` of course. I settled on `EXTENSIONSHORTNAME_symfony_HOOK` arbitrarily - normally I wouldn't add to the global namespace but point to a class method.

MikeyMJCO:
Great work @christianwach thanks for the effort - I'm in favour of the general idea behind this - with the specific implementation up for grabs/discussion :-) I think this is essentially a Proof of Concept for me at this stage.

christianwach:
> we're no longer handing off to the hook implementation in core right

Actually we are :-) The event that triggers the callback *is the hook*. We're just bypassing the clunky `invokeViaUF()` method, which you can [see in all its glory](https://github.com/civicrm/civicrm-core/blob/master/CRM/Utils/Hook.php#L238-L306) in `CRM_Utils_Hook::runHooks()`.

christianwach:
> I think this is essentially a Proof of Concept for me at this stage.

Yup, for me too. Thanks for taking the bait and your thoughtful responses.

JoeMurray:
@totten this kind of architectural discussion deserves some attention from you. 

Somewhat random thoughts because editing is hard on my cellphone.

I like in the abstract using a listener pattern, which is not quite as general as an observer pattern. Others can wrangle important technical conventions around them. I would encourage a review of other PHP projects for their syntax, conventions, etc especially the big CMSes and other PHP frameworks (though obviously Symfony is our preferred one).

Hooking pattern as per wikipedia entry is similar but in our Drupal like implementation does not really allow dynamic deregistration except through disabling the extension. 

We have many hooks that listen to event streams that they filter on themselves, eg selecting for a specific form to act on. Slightly inefficient. We adopted hooks from Drupal in Civi because our dev community at the time was familiar with it.

I would be concerned about maintaining two programming patterns that overlap a bunch in terms of the problems they solve. It will be adding to the already significant dev learning curve. This is only one consideration.

Can we figure out a way to achieve the desired objective here without introducing a new pattern is my first question?

If we approve this then I will want to see a new section of dev documentation, including criteria for when it is accrptable/appropriate to add a listener to vore. That should explain when to use a hook vs when to use a listener.

MegaphoneJon:
WRT the concerns @JoeMurray raises, Nicolas raised similar issues when this pattern first became available last year. See https://github.com/civicrm/org.civicrm.flexmailer/issues/1 and https://github.com/civicrm/civicrm-core/pull/9949.

While I think there are valid concerns here, this is definitely the pattern that @totten supports because he wrote the original implementation.

JoeMurray:
If totten introduced pattern then my two remaining concerns are 1) is it really needed here? I think it likely is advantageous and believe there be consensus on that, but still want to ask.
2) can we have some standard docs like for hooks, not as fancy as for api? Especially for hooks / listeners we should try to have simultaneous PRs for docs and interface changes.

eileenmcnaughton:
https://docs.civicrm.org/dev/en/latest/hooks/setup/symfony/

JoeMurray:
How embarassing - missed reading that completely and I've unaware of our longstanding support for Symfony listeners. 

So most hooks are able to be documented once and used via CMS or Symfony 'instantiations'. 

Would be nice to have some guidance on when to use Symfony approach, eg when you want to end listening, etc.

If new hook 
hook_shoreditch_civicrm_config is documented then I'm okay with this.

BTW it would be good to standardize documentation location for hooks defined by core extensions. I think they should go in with core ones and have different naming convention like this one.

christianwach:
> If new hook `hook_shoreditch_civicrm_config` is documented then I'm okay with this.

I'm not wedded to the naming convention that I - ahem - made up on the spot for this hook. I'd appreciate advice from someone more familiar with such matters. 

My main goal here was to apply a pattern that is very common in the WordPress ecosystem to CiviCRM extensions. Many WordPress plugins offer hooks and indeed many are built on the hooks that they themselves provide. Some kind of "hey there, I've initialised!" hook is very common -- the CiviCRM WordPress plugin, for example, has [had this for some time](https://github.com/civicrm/civicrm-wordpress/blob/master/civicrm.php#L695) -- so that when I write code which is dependent on CiviCRM being initialised, I often trigger it by listening for this hook. CiviCRM plugin deactivated? No problem, my code never runs.

It would be a nice convention for CiviCRM extensions to implement - though it might be better to enforce it (if that's desirable) by shifting the `hook_shoreditch_civicrm_config` dispatch code to the auto-generated `shoreditch.civix.php` file and moving the call to `_shoreditch_civix_civicrm_config()` to the end of `shoreditch_civix_civicrm_config()`. The naming convention could then be formalised in [Civix](https://github.com/totten/civix/):

```php
/**
 * Implements hook_civicrm_config().
 *
 * @link http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_config
 */
function shoreditch_civicrm_config(&$config) {
  // Add listeners for CiviCRM hooks that might need altering by other scripts
  Civi::service('dispatcher')->addListener('hook_civicrm_coreResourceList', 'shoreditch_symfony_civicrm_coreResourceList', -100);
  Civi::service('dispatcher')->addListener('hook_civicrm_buildForm', 'shoreditch_symfony_civicrm_buildForm', -100);
  Civi::service('dispatcher')->addListener('hook_civicrm_pageRun', 'shoreditch_symfony_civicrm_pageRun', -100);

  // Lastly, hand off to Civix
  _shoreditch_civix_civicrm_config($config);
}
```

@totten Would this be worth pursuing or is it easier to let extensions do their own thing?

christianwach:
One thing that has surprised me about the Symfony Event Dispatcher is that when an event is in progress, no changes can be made to the listeners of that event with a higher priority. So, for example:

```php
class Cannot_Unhook_Seemingly {

	/**
	 * Register hooks.
	 */
	public function register_hooks() {
		add_action( 'civicrm_config', array( $this, 'config_callback' ), 10, 1 );
	}

	/**
	 * Implements hook_civicrm_config().
	 *
	 * @param array $config The config settings.
	 */
	public function config_callback( &$config ) {

		// listen for Shoreditch config complete
		Civi::service('dispatcher')->addListener(
			'hook_shoreditch_civicrm_config',
			array( $this, 'fires_first' ),
			-100 // priority
		);

		// add a second listener for Shoreditch config complete
		Civi::service('dispatcher')->addListener(
			'hook_shoreditch_civicrm_config',
			array( $this, 'fires_second' ),
			-200 // higher priority
		);

	}

	/**
	 * Fire first when Shoreditch is configured.
	 *
	 * @param \Civi\Core\Event\GenericHookEvent $event The event object.
	 * @param str $hook The name of the hook currently being run.
	 */
	public function fires_first( $event, $hook ) {

		// try and remove a later callback
		Civi::service('dispatcher')->removeListener(
			'hook_shoreditch_civicrm_config',
			array( $this, 'fires_second' )
		);

	}
	
	/**
	 * Fire second when Shoreditch is configured.
	 *
	 * @param \Civi\Core\Event\GenericHookEvent $event The event object.
	 * @param str $hook The name of the hook currently being run.
	 */
	public function fires_second( $event, $hook ) {
	
		// should not receive a callback - but does!

	}
	
} // class ends

// let's go
$my_test_object = new Cannot_Unhook_Seemingly();
$my_test_object->register_hooks();
```

This is common practice in WordPress -- though some of [the oddities took years to solve](https://core.trac.wordpress.org/ticket/17817) -- so I thought it worth noting here.

davialexandre:
Thanks for this @christianwach and everyone else! Having a way to not inject the custom css in some places can really be useful. Not only for Shoreditch, but for any other theme. For this reason, we think this should be something implemented in civicrm core, rather than each theme implementing it in their own way.

There's already a Gitlab issue with a proposal for this: https://lab.civicrm.org/dev/core/issues/378. So I think we should probably continue the discussion in there

christianwach:
Thanks for the link @davialexandre - good to see there's thinking around issues like this one. 

I think there's a misunderstanding though: this PR is not about addressing the inclusion (or not) of a custom CiviCRM or Shoreditch CSS file - it's about [the resources that Shoreditch loads](https://github.com/civicrm/org.civicrm.shoreditch/blob/master/shoreditch.php#L121-L132) and the inability of external code to prevent that from happening, whether in part or at all.

In practical, real-world terms the primary target of this PR is to _prevent `bootstrap.css` from loading_ because it does not play well with CMS themes that are themselves Bootstrap-based. Yes, one could try and solve this with the addition of yet another option but I would argue that it is almost impossible to predict all the circumstances when this file should or should not be included. The same goes for any other resource loaded by Shoreditch - do we really want an option for each one?

The switch to Symfony event listeners would mean that you, as Shoreditch extension developers, don't have to concern yourselves with predicting all those circumstances, figuring out and maintaining all that logic (not to mention all the inevitable edge cases and support requests) because that effort is outsourced to site, module, plugin and extension devs who are themselves able to decide what to do in their particular circumstances.

As it stands, this PR enables those advantages without affecting existing behaviour in any way.

deb1990:
I like the general idea, but as @davialexandre said, its better to implement this on core, which will be globally applicable for all extensions. Otherwise we would have to move all hooks to symfony implementation, which will get messy in the long run. So we would need a centralised pattern from core to deregister hooks. So I suggest to open a discussion on core.

Closing as its inactive for a long time.

PR #320, opened 2018-09-26
SHOR-93: Exclude dist files from PRs + gulp release task 

AkA84:
I had to rollback #315 due to some side effects of the solution I didn't fully consider yet at the time. This PR has the content of the original one (reference #315 for details) + additional changes that are documented below

### post-checkout hook
This hook will trigger `gulp sass` if it will detect changes to the source code (see [here](https://git-scm.com/docs/githooks#_post_checkout) for an explanation on when the hook is invoked).

```bash
SRC_PRESENT=$(git diff @{-1} HEAD --exit-code --name-only -- *.scss)

if [ -n "$SRC_PRESENT" ]
then
  echo -e "${YELLOW}Changes in *.scss files detected, rebuilding dist files...${RESET_COLOR}"
  npx gulp sass
fi
```

Given how "aggressive" the hook is, it is _not_ installed automatically by husky on `postinstall` but instead kept optional and it's up to the individual dev to decide whether to use it or not.

The particular reason the hook was developed was to support an automated setup where a branch is deployed (via `git checkout -b branch-to-test origin/branch-to-test`) on a test site for QA purpose

### Removed extensions from hook files
The hook files have been stripped of their extension (ie `post-checkout.sh` -> `post-checkout`) given that is the standard filename format git expects to be used

### Upgraded to husky v1.0
Husky just released version [`v1.0.0`](https://github.com/typicode/husky/releases/tag/v1.0.0), so made sense to upgrade to it.

One good thing about the new version is that, if you have a custom git hook already present in `.git/hooks/`, husky will not replace it (something instead would happen with the 0.x versions). This works well in case one wants to manually install the post-checkout hook described above

### Automatic ignore of css files on `postinstall`
```json
"scripts": {
  "postinstall": "npx gulp sass && git update-index --skip-worktree css/*"
}
```
The command `git update-index --skip-worktree css/*` automatically puts the .css files in ignore on the local repository, so that the developer doesn't have them in the way while staging and committing source code

(this can be undone by running `git update-index --no-skip-worktree css/*`)

As a consequence, the `gulp release` task had to be enhanced with an additional check: if the local repo of the releaser is ignoring the .css files, then the release process fails (as it can't add those files to the release commit if they are ignored)
```js
gulp.task('release:worktree-check', () => {
  return git.raw([
    'ls-files',
    '-v'
  ])
    .then(files => {
      return !!files.split('\n').filter(file => {
        return file.indexOf('S css/') === 0;
      }).length;
    })
    .then(areMinFilesSkipped => {
      if (areMinFilesSkipped) {
        throw new PluginError('release:worktree-check', {
          message: `The .css files are currently ignored in your local repo, unable to create the release commit.
  Please run \`git update-index --no-skip-worktree css/*\` before running the release command again.`,
          showProperties: false
        });
      }
    });
});
```
 
### Updated documentation
The documentation has been updated to reflect the addition of the new hook and the automatic ignore of .css files

### Distributed configurations in .rc files
The `package.json` was getting too cluttered with different configuration objects for modules, linters, etc (see [here](https://github.com/civicrm/org.civicrm.shoreditch/blob/837e0a56e44101f97e33bc202bf728d4ba323f1f/package.json)). So wherever possible those objects were distributed across different .rc files: .huskyrc, .lintstagedrc, .stylelintrc. The only configuration object left is the one for "semistandard", who doesn't support .rc files


.githooks/post-checkout
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
deb1990: Can we add documentation in the file, to explain what this does.

.githooks/post-checkout
```diff
@@ -0,0 +1,12 @@
...
+
+YELLOW='\033[0;33m'
+RESET_COLOR='\033[0m'
+
+SRC_PRESENT=$(git diff @{-1} HEAD --exit-code --name-only -- *.scss)
```
deb1990: can we change this to `IS_SCSS_UPDATED` or something similar

.githooks/post-checkout
AkA84: @deb1990 the files are pretty straightforward and they are fully documented in the CODING.md file, i think any additional comment here will be redundant

.githooks/post-checkout
AkA84: i've used more descriptive names

PR #324, opened 2018-10-08
BASW-269: Misc improvements and code cleanup

AkA84:
This contains several style improvements and a general code clean-up of the css related to form fields and [select2](https://select2.org/)

### Before
The sub-tabs in the Contribution tab (introduced in [this PR](https://github.com/civicrm/civicrm-core/pull/11956)) were not styled correctly, due to the ad-hoc style of the contact summary's main tabs being too generic

tabs-before

### After
Cleaning up, fixing linter issues (wherever possible) and narrowing the focus of the selectors in the [civicrm/contact/_tabs.scss](https://github.com/civicrm/org.civicrm.shoreditch/blob/master/scss/civicrm/contact/_tabs.scss) partial (with some additional fixes in [jquery/overrides/_tabs.scss](https://github.com/civicrm/org.civicrm.shoreditch/blob/master/scss/jquery/overrides/_tabs.scss) as well) fixed the issue

tabs-after

## Form fields
The form fields were styled by importing the bootstrap partials and then by `@extend`ing bootstrap classes. This kind of "brute force" approach worked, but it lead to the stylesheet having very large css selectors that were extremely hard to navigate, maintain, and extend. On top of that, there were many ad-hoc styles whose only purpose was to alter the original layout of the some of forms in core screens.

This PR is a first step toward simplifying the css styles for forms, by:
* removing the heavy reliance on `@extend`. The form elements are retaining the same bootstrap-like look and feel, with the difference that now the style that achieves that look-and-feel is fine tuned to have as few as css rules as possible.
* removing any unnecessary layout-related ad-hoc styles

The above results in a much lighter stylesheet, much simpler selectors that are easier to maintain and extend, and forms that retain the original layout. An example can be seen below

fields-width-diff
_(when the scrubber swipes left, the "after" screenshot is revealed. When it swipes right, the "before" screenshot is revealed)_

## Select 2
The styles for the [select2](https://select2.org) dropdown underwent the same treatment described for the form fields

* simplified the code
* removed layout-related ad-hoc styles

select2-diff
_(when the scrubber swipes left, the "after" screenshot is revealed. When it swipes right, the "before" screenshot is revealed)_

Notice that now the correct icon, the magnifying glass, is displayed for the `crm-ajax-select` version of select2 (that is, the version that fetches the data via AJAX requests) instead of always showing the caret for every version. Additionally the style for the `<a>` element inside the select is maintained, just like in the original civi theme


scss/civicrm/common/_backgrounds.scss
```diff
@@ -21,7 +21,7 @@
     .crm-form-block {
       @extend %block-with-shadow;
       border-radius: 0;
-      margin: 0;
+      margin-bottom: 15px;
```
igorpavlov: Why do we add exactly `15px`? Is there a variable for that?

If you have a look at `scss/bootstrap/overrides/_variables.scss` and search for `15px` you will see this value is actually used in a lot of places. It seems to me like this value is pretty universal and it should be defined in a separate variable like `$standard-vertical-grid-gap` and then using it everywhere when you use `15px` because it is a generic grid design decision.

scss/civicrm/contact/_tabs.scss
```diff
@@ -1,194 +1,195 @@
-/* stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type, selector-max-id */
+/* stylelint-disable selector-max-id */
```
igorpavlov: Is this against CODING.md rule when we say that it is a must to have all rules disabled? I may understand why you do this, but should we then say in CODING.md that `if you *can* enable some rules, please go ahead`?

scss/civicrm/contact/_tabs.scss
```diff
@@ -1,194 +1,195 @@
...
 
-    .crm-contact-tabs-list {
-      background: $crm-white;
-      border-radius: 0;
-      box-shadow: $box-shadow;
-      flex-shrink: 0;
-      margin: 0;
-      padding: 0;
-      width: 200px;
+  // override of jquery-ui.css that creates weird white background
```
igorpavlov: Let's remove any opinion based statements from comments and stay cold-blooded:

`// Override a white background rule made by jquery-ui.css`

scss/civicrm/contact/_tabs.scss
```diff
@@ -1,194 +1,195 @@
...
-        em {
-          color: $brand-primary !important;
-        }
-      }
+  a {
+    box-sizing: border-box;
+    display: block;
+    font-size: $font-size-small !important;
+    height: 100%;
+    padding: 15px !important;
```
igorpavlov: Same as above about `15px`.

scss/civicrm/contact/_tabs.scss
```diff
@@ -1,194 +1,195 @@
...
+      color: $gray-darker !important;
+      cursor: default;
+    }
   }
 
-  #ui-id-11 {
-    overflow: auto;
+  ~ .ui-tabs-panel {
+    flex-grow: 1;
+    padding: 20px !important;
```
igorpavlov: Same as `15px` suggestion, but this time for `20px`

scss/civicrm/contact/_tabs.scss
```diff
@@ -1,194 +1,195 @@
...
+      // Necessary to override the id-based namespaces of T&A and JC
+      // (as they still use their own copy of Bootstrap)
+      max-width: $container-large-desktop !important;
     }
   }
 
-  .ui-dialog {
-    background: $crm-white;
+  // Necessary because of the !important defined in
+  // https://github.com/civicrm/civicrm-core/blob/cc38fb8ddcbbc8b60679504dd93522df852c48bf/css/civicrm.css#L1616-L1619
```
igorpavlov: `// @see https://github.com/civicrm/civicrm-core/blob/cc38fb8ddcbbc8b60679504dd93522df852c48bf/css/civicrm.css#L1616-L1619`

scss/civicrm/contact/_tabs.scss
```diff
@@ -1,194 +1,195 @@
...
+  .ui-tabs-anchor {
+    padding: 15px !important;
```
igorpavlov: Same as above about `15px`

scss/civicrm/contact/pages/_batch.scss
```diff
@@ -221,14 +192,11 @@
         }
 
         .crm-form-date-wrapper {
-          width: 370px;
+          display: inline-block;
+          width: 255px;
```
igorpavlov: This change is a bit worrying. Why it is exactly 255px, can't it be calculated somehow from other variables? At least, let's move that to a separate variable.

scss/civicrm/contact/pages/_contributions.scss
```diff
@@ -596,6 +538,10 @@
       @include expandable-table;
     }
   }
+
+  .crm-contact-contribute-recur-active {
+    margin-bottom: 13px;
```
igorpavlov: Why exactly `13px`?

scss/civicrm/search/pages/_advanced-search.scss
```diff
@@ -1,95 +1,16 @@
...
-      margin-bottom: 20px;
+  .adv-search-top-submit {
+    .crm-button {
+      float: right !important;
     }
   }
 
   .form-layout {
     .crm-i-button {
+      margin-left: 20px;
```
igorpavlov: Same as other `20px` comment

scss/jquery/overrides/_select2.scss
```diff
@@ -0,0 +1,246 @@
...
+.select2-container {
+  // The input field overlay
+  .select2-choice {
+    background: $crm-white;
+    border-color: $crm-grayblue-darker;
+    border-radius: $border-radius-base;
+    box-shadow: $crm-form-control-shadow;
+    box-sizing: border-box;
+    height: $input-height-base;
+    line-height: 27px;
```
igorpavlov: Why `27px`?

scss/jquery/overrides/_select2.scss
```diff
@@ -0,0 +1,246 @@
...
+
+    &:focus {
+      outline: 0;
+    }
+
+    // The X icon to deselect the option
+    abbr {
+      background-image: none !important;
+      right: 30px;
```
igorpavlov: Something similar to `15px` and `20px` concerns.

scss/jquery/overrides/_select2.scss
```diff
@@ -0,0 +1,246 @@
...
+        font-size: $font-size-base;
+      }
+    }
+
+    .select2-arrow {
+      background: $crm-white;
+      border-left-color: $crm-grayblue-darker;
+      border-radius: 0 $border-radius-base $border-radius-base 0;
+      text-align: center;
+      width: 26px;
```
igorpavlov: Why `26px`?

scss/jquery/overrides/_select2.scss
```diff
@@ -0,0 +1,246 @@
...
+    }
+  }
+
+  // Wrapper
+  .select2-choices {
+    background: $crm-white;
+    border-color: $crm-grayblue-darker;
+    border-radius: $border-radius-base;
+    box-shadow: $crm-form-control-shadow;
+    line-height: 27px;
```
igorpavlov: Same as above.

scss/jquery/overrides/_select2.scss
```diff
@@ -0,0 +1,246 @@
...
+      top: 1px;
+    }
+
+    // The selected option
+    .select2-search-choice {
+      background: none;
+      border-color: $brand-primary;
+      border-radius: $border-radius-base;
+      font-size: $font-size-base;
+      padding: 3px 25px 3px 5px;
```
igorpavlov: Why such a value? Can it be calculated from some other variables already defined?

scss/jquery/overrides/_select2.scss
```diff
@@ -0,0 +1,246 @@
...
+
+      &.select2-search-choice-focus {
+        background: $gray-light;
+      }
+
+      /* stylelint-disable selector-max-compound-selectors, max-nesting-depth */
+      .select2-search-choice-close {
+        background: none !important;
+        left: auto;
+        right: 4px;
```
igorpavlov: Why `4px`?

scss/jquery/overrides/_select2.scss
```diff
@@ -0,0 +1,246 @@
...
+    .select2-search-field {
+      /* stylelint-disable selector-max-compound-selectors */
+      input {
+        @include open-sans-font-family();
+        // These !important are necessary to counteract Bootstrap's base style
+        // when select2 is used inside the bootstrap namespace
+        font-size: $font-size-base !important;
+        line-height: normal !important;
+        margin: 0;
+        padding: 4px 4px 6px 8px;
```
igorpavlov: Same as above.

scss/jquery/overrides/_select2.scss
```diff
@@ -0,0 +1,246 @@
...
+      &.select2-search-choice-focus {
+        background: $gray-light;
+      }
+
+      /* stylelint-disable selector-max-compound-selectors, max-nesting-depth */
+      .select2-search-choice-close {
+        background: none !important;
+        left: auto;
+        right: 4px;
+        top: 4px;
```
igorpavlov: Same as above.

scss/jquery/overrides/_select2.scss
```diff
@@ -0,0 +1,246 @@
...
+        // These !important are necessary to counteract Bootstrap's base style
+        // when select2 is used inside the bootstrap namespace
+        font-size: $font-size-base !important;
+        line-height: normal !important;
+        margin: 0;
+        padding: 4px 4px 6px 8px;
+
+        // Spinner icon that appears when fetching data via ajax
+        &.select2-active { /* stylelint-disable-line max-nesting-depth, selector-no-qualifying-type */
+          background-position: center right 20px !important;
```
igorpavlov: Same as above (20px)

scss/jquery/overrides/_select2.scss
```diff
@@ -0,0 +1,246 @@
...
+      }
+      /* stylelint-enable selector-max-compound-selectors */
+    }
+
+    // The list of items that match the search
+    .select2-results {
+      /* stylelint-disable selector-max-compound-selectors */
+      li {
+        color: $gray-dark;
+        padding: 5px;
```
igorpavlov: Same as above: 10px 20px etc...

scss/jquery/overrides/_select2.scss
```diff
@@ -0,0 +1,246 @@
...
+      .select2-highlighted {
+        background: $gray-light;
+
+        p { /* stylelint-disable-line max-nesting-depth */
+          color: $gray-dark;
+        }
+      }
+
+      .select2-result-label {
+        padding: 5px 10px;
```
igorpavlov: Same as above.

scss/jquery/overrides/_select2.scss
```diff
@@ -0,0 +1,246 @@
...
+
+  // Search text input field
+  .select2-search {
+    &::before {
+      @include fa-icon();
+      color: $gray-darker;
+      content: $fa-var-search;
+      font-size: $font-size-base;
+      position: absolute;
+      right: 10px;
```
igorpavlov: Same as above.

scss/jquery/overrides/_select2.scss
```diff
@@ -0,0 +1,246 @@
...
+      top: 10px;
+    }
+
+    input {
+      @include open-sans-font-family();
+      background-image: none !important;
+      border-color: $input-border-focus;
+      line-height: 20px;
```
igorpavlov: Same as above.

scss/bootstrap/overrides/crm/_form-text.scss
```diff
@@ -1,8 +1,5 @@
-// Removes fixed with for text inputs when they have the size attribute:
+// Bootstrap applies border-box to everything, so the height
+// of .crm-form-text needs to be adjusted accordingly
 .crm-form-text {
-
-  &[size] {
```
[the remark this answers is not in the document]
AkA84: that's the old style, not the style contained in this branch

scss/civicrm/administer/contribute/_contribute.scss
```diff
@@ -112,10 +108,6 @@
   }
 
   .crm-contribution-contributionpage-premium-form-block {
-    textarea + br {
```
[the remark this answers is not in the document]
AkA84: rather that we were trying to fix the markup via css, instead of just submitting a core patch that removed that `<br>`

scss/civicrm/contact/_tabs.scss
AkA84: yes i actually regret adding that get-out-of-jail-free line in CODING.md. Of course instead it should say that it's ok to disable some rules, _if necessary_. There are plans to overhaul the documentation in the readme, i'll fix that line then (i need to come up with a good wording and maybe some examples, etc)

scss/civicrm/contact/pages/_batch.scss
AkA84: Will leave a comment

scss/jquery/overrides/_select2.scss
AkA84: The line-height needs to be lower than the actual height of the element, otherwise the text inside the select2 will not be properly centered vertically

line-height

_(with `line-height: $input-height-base;`)_

AkA84: I can do
```scss
line-height: $input-height-base - 3;
```
what do you think

scss/jquery/overrides/_select2.scss
AkA84: It's the width of `.select2-arrow` + some additional margin

scss/jquery/overrides/_select2.scss
AkA84: No particular reason, is just how much the arrow needs to be wide according to the designs

scss/jquery/overrides/_select2.scss
AkA84: created a variable

scss/jquery/overrides/_select2.scss
AkA84: created a variable

scss/jquery/overrides/_select2.scss
AkA84: created a variable

scss/jquery/overrides/_select2.scss
AkA84: created a variable

scss/jquery/overrides/_select2.scss
AkA84: not really, it's an adjustment of the civicrm style
```css
.crm-container .select2-container-multi .select2-choices .select2-search-choice {
  padding: 2px 5px 2px 18px;
}
```
which is an adjustment of the original select2 style
```css
.select2-container-multi .select2-choices .select2-search-choice {
 padding: 3px 5px 3px 18px;
}
```

scss/jquery/overrides/_select2.scss
AkA84: Same as before, an adjustment of the civicrm style
```css
.crm-container .select2-search-choice-close {
  top: 2px;
}
```
which is in ad adjustment of the original select2 style
```css
.select2-search-choice-close {
  top: 4px;
}
```

scss/jquery/overrides/_select2.scss
AkA84: Needed to have the correct height and padding that matches to the single version of select2

scss/jquery/overrides/_select2.scss
AkA84: Adjustment of the civicrm's style, same as above

scss/jquery/overrides/_select2.scss
AkA84: It's simply that the search items have some paddings instead of the default style which has none

scss/jquery/overrides/_select2.scss
AkA84: adjusting the original padding, nothing more

scss/jquery/overrides/_select2.scss
AkA84: it's simply to align the search icon in the search field

icon

scss/jquery/overrides/_select2.scss
```diff
@@ -17,7 +22,7 @@
     // The X icon to deselect the option
     abbr {
       background-image: none !important;
-      right: 30px;
+      right: $crm-jquery-select2-arrow-width + 4;
```
igorpavlov: What is `4px`?

PR #343, opened 2018-12-13
BASW-340: Fixed top margin for body with adminimal_admin_menu module class

ivan-compucorp:
Civicrm menu was overlapping the header because adminimal_admin_menu module breakes body top margin. Now it fixed.

Before:
image

After:
image

### DIFF
--- a/scss/civicrm/common/_base.scss
+++ b/scss/civicrm/common/_base.scss
@@ -11,10 +11,13 @@
         margin-top: $crm-main-menu-height !important;
       }
 
+      // These rules are needed to fix the visual issue produced by Adminimal Menu module: it removes body top-margin but adds `body::before{height: 29px;...}`.
+      // That causes Civicrm menu overlap the header which looks badly.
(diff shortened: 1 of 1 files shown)


igorpavlov:
@ivan-compucorp could you please specify which exact rule in the module breaks the layout?

ivan-compucorp:
@igorpavlov afair it was something like that (comes from adminimal module):
`body.adminimal-menu:after{height: 29px}`

scss/civicrm/common/_base.scss
```diff
@@ -11,6 +11,14 @@
         margin-top: $crm-main-menu-height !important;
       }
 
+      &.admin-menu.adminimal-menu {
```
igorpavlov: Please leave a comment here, smth like:

```
// These rules are needed to fix the visual issue produced by Adminimal Menu module.
// It adds an additional gap between....
// @see <url-to-module>
```

ivan-compucorp: @igorpavlov fixed, check please.

AkA84:
thanks @colemanw , we agreed to close the PR also given what you suggested

PR #346, opened 2019-01-09
SHOR-105: Fix Contact Custom Data Tab

reneolivo:
## Overview
This PR fixes the custom data data displayed on the contact details. The table is not expanded properly due to an outdated CSS rule. Also, the alignment of the delete action button has been fixed.

## Before
screen shot 2019-01-08 at 10 36 22 pm

## After
screen shot 2019-01-10 at 10 52 58 am

## How to reproduce
* Create a custom field group for contacts.
* Add new fields to that custom field group.
* Open any contact and view their details.
* Click on the tab with the name of the new custom field group.

## Technical Details

The problem is that the following rule:

```scss
table.crm-info-panel td.html-adjust {
  width: 100%;
}
```

Overrides the natural width of the column. The label column is `150px` and the value column would take up the rest, but if `width: 100%` is specified the column would try to take as much space as possible and break the label column's width.

This rule was first introduced in the following PR for an old CiviCRM version:
https://github.com/compucorp/civihr/pull/1598

Also, the alignment of the actions button was fixed by removing the padding from their container and removing the padding left of the first button:

```scss
.crm-submit-buttons {
  padding: 0;

  .crm-hover-button:first-child {
    padding-left: 0;
  }
}
```

Also, the icon for the button was aligned to the baseline so it's symmetrical in relation to its companion text:

```scss
.crm-submit-buttons 
  .crm-i {
    vertical-align: baseline;
  }
}
```

The border rules for the custom fields table have also been removed.

## Tests

Ran Backstop tests from the [Backstop Config](https://github.com/compucorp/backstopjs-config) repo. No changes were detected to the contact page tabs covered by the scenarios.

It's important to note that the custom field set tab was tested manually since there are no scenarios covering these type of tabs. A new scenario was not submitted to Backstop Config since this is very specific and requires multiple pre existing data.

### DIFF
--- a/scss/civicrm/contact/pages/_view-contact.scss
+++ b/scss/civicrm/contact/pages/_view-contact.scss
@@ -43,16 +43,16 @@
 
         table.crm-info-panel {
           border: 0;
-        }
 
-        table.crm-info-panel td {
-          color: $crm-copy;
-          vertical-align: middle;
-        }
+          td {
+            color: $crm-copy;
+            vertical-align: middle;
 
-        table.crm-info-panel td.label {
-          color: $gray-darker;
-          font-weight: $crm-font-weight-h1;
+            &.label {
+              color: $gray-darker;
+              font-weight: $crm-font-weight-h1;
+            }
+          }
         }
       }
 
--- a/scss/civicrm/contact/pages/_view-contact.scss
+++ b/scss/civicrm/contact/pages/_view-contact.scss
@@ -35,6 +35,10 @@
           .crm-hover-button:first-child {
             padding-left: 0;
           }
+
+          .crm-i {
+            vertical-align: baseline;
+          }
         }
 
         table.crm-info-panel {
--- a/scss/civicrm/contact/pages/_view-contact.scss
+++ b/scss/civicrm/contact/pages/_view-contact.scss
@@ -29,6 +29,14 @@
           border-radius: 0 0 $panel-border-radius $panel-border-radius !important;
         }
 
+        .crm-submit-buttons {
+          padding: 0;
+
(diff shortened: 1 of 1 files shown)


deb1990:
@reneolivo Code looks ok, did you run Backstop tests?

jamienovick:
Do we need the lines and also the delete button looks misaligned? Also the spacing seems a bit off. Can we make it look a bit more sensible?

reneolivo:
@jamienovick thanks for noticing. Fixed the button alignment and also the icon alignment.

reneolivo:
@deb1990 the icon is properly aligned. It looks like that because Font Awesome's trash icon is a bit lower than other icons:

**How it is now:**
screen shot 2019-01-10 at 10 46 56 am

**How it would look if we move the icons by 1px:**
screen shot 2019-01-10 at 10 47 10 am

I could add the `1px` rule for the trash icon only, but that would be playing favourites and it's something that might break if Font Awesome changes the dimensions of that icon.

scss/civicrm/contact/pages/_view-contact.scss
```diff
@@ -29,13 +29,20 @@
...
+          .crm-hover-button:first-child {
+            padding-left: 0;
           }
+
+          .crm-i {
+            vertical-align: baseline;
+          }
+        }
+
+        table.crm-info-panel {
```
deb1990: we should use nesting here
```
table.crm-info-panel {
  border: 0;

  td {
    color: $crm-copy;
    vertical-align: middle;

    &.label {
      color: $gray-darker;
      font-weight: $crm-font-weight-h1;
    }
  }
}
```

PR #347, opened 2019-01-15
SHOR-106: Fix margins for contact's participation custom fields

reneolivo:
## Overview
This PR fixes the following issues with custom fields for the contact's events tab:

* margins for the accordion title inside when creating a new event participation for the contact.
* The styles of the custom fields when viewing the participation.

### Creating a new event participation
admin example com shor 9600

### Viewing an existing participation
admin example com shor 9600 2

### Creating a new event participation
screen shot 2019-01-14 at 11 59 24 pm

### Viewing an existing participation
screen shot 2019-01-16 at 5 41 48 pm

### When creating a new event participation
Since the events form is a custom form with custom markup and css rules compared to those of other forms, the negative margin added here:
https://github.com/reneolivo/org.civicrm.shoreditch/blob/988ec9756af56a7cfa9940853ea2bc48f6cdace0/scss/civicrm/contact/pages/_events.scss#L281

Is removed for accordion wrappers:

```scss
.crm-accordion-wrapper {
  margin-left: 0;
  margin-right: 0;
}
```

### When viewing an existing event participation
The PR that first introduced the changes for the participation view modal is this one:
https://github.com/compucorp/civihr/pull/1639

The form has changed since then and the margin for the accordion headers was changed from `-40px` to `0px`:

```scss
[id*='Food_Preference'] {
  .crm-accordion-wrapper {
      margin-left: -40px;
      margin-right: -40px;
  }
}
```

Also, the `[id*='Food_Preference']` selector and all of its nested rules were changed to `.section-shown` since the former only assumes the fixes should be applied to the food preference custom fields, but the later fixes all issues for all custom fields.

## Tests

Backstops tests ran successfully.

### DIFF
--- a/scss/civicrm/contact/pages/_events.scss
+++ b/scss/civicrm/contact/pages/_events.scss
@@ -367,9 +367,11 @@
   }
 
   [id*='Food_Preference'] {
+    padding: 0;
+
     .crm-accordion-wrapper {
-      margin-left: -40px;
-      margin-right: -40px;
+      margin-left: 0;
+      margin-right: 0;
 
       .crm-accordion-body {
         padding: 20px 43px !important;
--- a/scss/civicrm/contact/pages/_events.scss
+++ b/scss/civicrm/contact/pages/_events.scss
@@ -281,6 +281,11 @@
       margin: 0 -20px;
     }
 
+    .crm-accordion-wrapper {
+      margin-left: 0;
+      margin-right: 0;
+    }
+
     .td {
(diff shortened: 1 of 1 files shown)


deb1990:
@reneolivo current solution looks good. but I found some additional issues

2019-01-15 at 4 41 pm

Can we fix the radio buttons as part of this PR or create separate tickets for them?

deb1990:
@reneolivo Approved, but please make sure backstop still passes after the latest change.

PR #352, opened 2019-01-28
BASW-376: Fix shadows of tables on Contribution tab

igorpavlov:
## Overview

This PR fixes the issue with shadows on the Contribution tab.

## Before

Notice overlayed shadows:
image

Notice the extra thick shadow on the table header:
image

## After

image

image

## Technical description

We add a mixin that supports this layout: H3 + Table in a Wrapper. We remove shadows from tables and headers and apply a solid shadow on the wrapper.

```css
@mixin civicrm-table-with-header() {
  box-shadow: $box-shadow;

  h3,
  table {
    box-shadow: none !important;
  }

  .columnheader {
    border: 0 !important;
    border-bottom: 1px solid $crm-grayblue-dark !important;
  }
}
```

## Comments

It was considered to:

1) find a universal style, but unfortunately there is no distinctive and reliable way to apply needed styles to only needed table due to inconsistency of markup.
2) change the markup - this is too complicated and will require changing markup in several extensions and also core.

## Tests

All changes in this PR are isolated to specific components/extensions, so manual tests were done. BackstopJS was run as well.

### DIFF
--- a/scss/civicrm/contact/pages/_contributions.scss
+++ b/scss/civicrm/contact/pages/_contributions.scss
@@ -543,6 +543,10 @@
   .crm-contact-contribute-recur-active {
     margin-bottom: 13px;
   }
+
+  .crm-contact-contribute-recur {
+    @include civicrm-table-with-header;
+  }
 }
 
 .CRM_Contribute_Form_Contribution {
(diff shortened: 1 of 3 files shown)


jamienovick:
@reneolivo @igorpavlov Not for this ticket perhaps but we also need a solution for the new tab between contributions and recurring contributions...

PR #356, opened 2019-02-08
Remove duplicate icons in tabs

### LINKED ISSUE #340: Remove icons from contact summary tabs
As of CiviCRM 5.8 the contact summary tabs have icons (based on the [recent styleguide update](https://github.com/civicrm/org.civicrm.styleguide/pull/22)). This will cause double icons unless they are removed from Shoreditch.

em-piguet:
This fixes #340 the double icon in the tabs by removing useless code.

## Before

image

## After
image


colemanw:
@AkA84 after merging this we should also bump the version requirement to at least 5.8.

PR #359, opened 2019-02-08
SHORE-341: fix quicksearch

### LINKED ISSUE #341: Quicksearch results show offscreen - top of page not view port.
To replicate:

Navigate to a civicrm screen which has enough content such there is enough to require a small amount of scrolling - 
This can be achieved from: /civicrm/admin?reset=1
Scroll down the page.
Type a search into the quicksearch bar. Notice that the Civicrm Logo will spin as the search is running then stop after it's finished - however no results are displayed.
Scroll to the top of the page without the form loosing focus (I can achieve this via mousewheel as long as you don't click as as soon as the input looses focus the results disappear.) 
You will see the results displayed at the top of the screen.

I've tested this behaviour on the below:
Shoreditch
Tested on:
 Civicrm 5.8.beta1 Drupal 7.60,  Shoreditch 0.1-alpha29User, 
 Civicrm 5.8.beta1 Drupal 7.60,  Shoreditch 0.1-alpha28
 Civicrm 5.3.1 Drupal 7.60,  Shoreditch 0.1-alpha28

I also check without using shoreditch  - on 5.8.beta1 and confirmed that this is not the behaviour without shoreditch - it doesn't appear to be a civicrm core issue.
quicksearch_position_cropped

em-piguet:
## Overview
This one fix #341 the position of the QuickMenu results 
## Test
Ok in Chrome and Firefox Mac (should be ok on others)

### DIFF
--- a/scss/civicrm/main-menu/_main-menu.scss
+++ b/scss/civicrm/main-menu/_main-menu.scss
@@ -88,11 +88,17 @@
 
   .menu-div {
     margin-top: 0 !important;
+    position: fixed;
   }
 
   .menu-item {
     font-size: $font-size-base;
-    padding: 6px 80px 6px 24px;
+    padding: 6px 24px;
+    width: 194px;
+
+    input {
+      margin-right: 8px;
+    }
   }
 
   .menu-item-arrow {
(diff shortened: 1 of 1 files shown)


colemanw:
This is superceded by #364

PR #360, opened 2019-02-13
BASW-379: Fix CiviCRM action item icons

igorpavlov:
## Overview

This PR fixes the issue with actions icons in CiviCRM.

## Before

image

## After

image

## Technical description

The problem is at *scss/civicrm/common/_buttons.scss*:

```
...
.crm-hover-button,
... {
  @extend %btn;
}
```

The `%btn` changes the font to `OpenSans` while it should be kept as `FontAwesome` but only for icons, so:

```
.crm-hover-button.crm-i {
  font-family: $font-family-fontawesome !important;
  // !important is needed because %btn contains !important too
}
```

is a safe way to keep the font correct and do not produce regressions.

## Tests

This is an extremely minor change, so only **manual** tests were performed to save time.

### DIFF
--- a/scss/civicrm/common/_buttons.scss
+++ b/scss/civicrm/common/_buttons.scss
@@ -189,6 +189,10 @@ $button-border-radius: 2px;
     @extend %btn;
   }
 
+  .crm-hover-button.crm-i {
+    font-family: $font-family-fontawesome !important;
+  }
+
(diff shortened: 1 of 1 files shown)


monishdeb:
Tested working fine. Based on two approval I am merging this PR

PR #363, opened 2019-02-18
BASW-377: Fix styles in new contribution modal

igorpavlov:
## Overview

This PR fixes appearance of contents inside modals. In a nutshell, it removes "per-page" code applied to "fix" the issues with the content, then applies global fixes, then applies _some_ "per-page" code where fixing globally is not viable.

## Before

The Recurring Contributions "view" modal looks totally broken:

image

image

## After

image

image

## More information

This PR applies new global styling for modals and their contents, thus a lot of modals now look slightly different, in a good way. This GIF is a very long one but it shows all modals and proves that global styling works:

modals

### Global modals style

At first, we created a new global style that applies to all modals and removed "per-page" styles related to modals, that fixed the initial screen that you see in Before/After sections and also other modals:

*scss/civicrm/common/_modals.scss*

```css
.ui-dialog-content {
  table {
    // ... global styles
```
There is just too much code to represent all in the snippet, so please refer to https://github.com/civicrm/org.civicrm.shoreditch/pull/363/files#diff-0508f2f5f60a92bc722462d514631b68 for more info.

### Message icon fix

```css
.messages.no-popup > .icon:first-of-type + p {
  margin-top: 0;
}
```
Makes the paragraph aligned with an icon.

#### Before

image

#### After

image

### Per-page fixes

```css
.crm-contact-contribute-contributions {
  margin: 0 -#{$crm-standard-gap};
  overflow-x: scroll;
  width: calc(100% + #{$crm-standard-gap * 2});
}
```
Fixes table in Memberships tab so it scrolls horizontally and does not get out of bounds of the modal.

### Nested accordion selector

This one may look weird, but there is no other way to identify this `<TD>`:

```css
&.form-layout,
  &.form-layout-compressed {
    > tbody > tr > td[colspan='2'] {
      padding: 0;
```
We need to reset paddings to fix nested accordions:

image

Unfortunately, the accordions are **inside** the `<TD>`s and there is no such CSS technique to select them, so we assume that if is `colspan=2` and it is not nested, then it is made for an accordion.

image

### Carets

Because we removed some "per-page" styles, we needed to align carets properly. Good news is that now we get rid of these ad-hocs like "top: 2px":

#### Before

image

#### After

image

## Tests

✅Manual Tests - passed (tested modals and random pages for accidental artifacts)
✅BackstopJS - passed

## QA round 1

We have a style for tables inside accordion body (needs negative margin):

```css
.crm-accordion-body {
    ...

    > table,
    > div > table {
      margin: 0 -#{$crm-standard-gap};
```
The problem is that for the data tables (class `.dataTable`) this should not be applied hence they already have wrappers that have negative margin.

image

So we need to exclude this wrapper:

```css
.crm-accordion-body {
    ...

    > table,
    > div:not(.dataTables_wrapper) > table { ...
```

image

Same with `.form-layout` and `.form-layout-compressed`.

Also, as per @jamienovick suggestion, we should not scroll tables horizontally in modals.

### Edit Event adhoc

The markup for the Edit Events is custom and needs to be treated separately:

image

```css
.crm-event-eventfees-form-block-payment_information fieldset,
  #payment_information fieldset,
  .crm-event-form-fee-block + fieldset {
    margin: 0 -#{$crm-standard-gap};
  }
```

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -4,6 +4,6 @@ In order to make sure that your work does not introduce regression issues or une
 
 The styleguide provides you with a page that showcases the entire base Bootstrap theme + custom modifiers in a single page, so that you can check how they look like in Shoreditch and how your changes affect the theme and its components.
 
-To run BackstopJS on the styleguide, follows the [instructions](https://github.com/compucorp/org.civicrm.styleguide/blob/staging/README.md#backstopjs) included in the extension. Make sure to always use BackstopJS on the styleguide whenever you work on `bootstrap.css`.
+To run BackstopJS on the styleguide, follows the [instructions](https://github.com/compucorp/org.civicrm.styleguide/blob/master/README.md#backstopjs) included in the extension. Make sure to always use BackstopJS on the styleguide whenever you work on `bootstrap.css`.
 
 To run BackstopJS on the existing core screens, you can temporarily use [this setup](https://github.com/compucorp/backstopjs-config). Of course this should be used only if you are introducing changes to `custom-civicrm.css`.
--- a/scss/civicrm/_variables.scss
+++ b/scss/civicrm/_variables.scss
@@ -32,6 +32,8 @@ $crm-min-lg-width: 1505px;
 $crm-sm-th-width: 200px;
 $crm-sm-input-width: 100px;
 
+$crm-table-label-column-width: 150px;
+
 $crm-status-bar-no-top: 0 20px 20px;
 $crm-status-bar-no-bottom: 20px 20px 0;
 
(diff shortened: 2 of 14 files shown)


jamienovick:
@igorpavlov Would it be possible where we have the horizontal scroll, to instead just have the width of the modal fixed to the max width of the elements within it? I think that would be neater?

igorpavlov:
@jamienovick Sounds like a good idea. Created a simple JSfiddle:

https://jsfiddle.net/igorpavlov/gvdk6oaz/4/
modal

So that's definitely possible. However, what to do with extremely wide tables that take more than the screen width?

PR #364, opened 2019-02-22
Update css for kam smartmenu

### LINKED ISSUE #342: Remove all menubar css and js
The soon-to-be-in-core [KAM extension](https://github.com/aydun/uk.squiffle.kam) natively implements basically all the things that Shoreditch was adding onto the menubar - lighter color scheme, robust markup, larger size, etc.

I'd recommend we remove *all* menubar-related css and JS from Shoreditch and start from scratch if there are any remaining tweaks to be done.

colemanw:
Fixes #342 

I understand there are minor style differences between KAM and the Shoreditch menu, notably the height (41px vs 60px) and the color (white on lightgrey vs white on white) however if we don't merge this Shoreditch will be broken so I suggest we think about style tweaks later and get this merged in sooner.


colemanw:
@AkA84 ping

colemanw:
@AkA84 @jamienovick I understand there are some discussions to be had about tweaking the style of the new KAM menu. But whether or not we do that, this PR is required. Please merge.

igorpavlov:
The PR looks clean to me. cc @AkA84

AkA84:
thanks @igorpavlov , but this PR is _not to be merged yet_, will do myself when we're ready for it

jamienovick:
@aka84 do you think we can get this styled as per Shoreditch also... Hate to be a pain but it is the small things that will make the difference.

colemanw:
@jamienovick I just want to warn you that changing the height of the menu bar would be quite difficult.
Because CiviCRM core does no use SCSS (maybe it should but that's a different conversation) there are many, many places in many core css files that do calculations based on the current height of 40px. If you grep the css files for "40px" or "80px" etc. you'll find an overwhelming number of results. Too many for Shoreditch to safely attempt to mess with IMO.

PR #368, opened 2019-02-25
SHOR-122: Style CiviRules extension

igorpavlov:
## Overview

This PR styles CiviRules extension pages.

## Before

_Please note, this page was just used as an example_

image

## After

image

### Block form mixin

We need this mixin to fix the layout of block forms.

```css
@mixin block-form() {
  // otherwise, fields are attached to the borders, which is ugly
  padding-bottom: $crm-standard-gap !important;
  padding-top: $crm-standard-gap !important;

  // otherwise, fields are attached to the borders, which is ugly
  .label,
  .content {
    padding-left: $crm-standard-gap;

  .content {
    padding-right: $crm-standard-gap;

  // no last margin needed
  .crm-section:last-of-type {
    margin-bottom: 0;

  // used in each row
  .clear {
    // This value is taken from the vanilla CiviCRM
    margin-bottom: 1em;

    &:last-of-type {
      margin-bottom: 0;

  // this is for other type of tables
  .form-layout {
    margin-top: $crm-standard-gap;

    tr td:first-of-type {
      padding-left: $crm-standard-gap;
```

### Table `.selector` mixin

This mixin is needed for two reasons:
1) style the results table
2) do not produce unexpected regressions

```css
@mixin table-selector() {
  // sometimes search tables have missing <TD>s
  background-color: $crm-white;

  thead th { // in civi we have other table header styles like this
    background-color: $gray-light !important;

  tbody tr:not(:last-of-type) {
    border-bottom: solid 1px $gray-light !important;

  td,
  th {
    border: 0 !important;
    line-height: 1.5em !important; // goes from civi styling
    // why not use a formula if we have a base variable?
    padding-bottom: #{$crm-standard-gap / 4 * 3} !important;
    padding-left: $crm-standard-gap !important;
    padding-top: #{$crm-standard-gap / 4 * 3} !important;

    &:last-of-type {
      padding-right: $crm-standard-gap !important;
```

#### Before

image

#### After

image

### Rules Extension styles

Please note, not all code is included, only the "tricky" places:

```css
// main selector for the extension
.page-civicrm-civirule {
  &.page-civicrm-civirule-form-rule {
    // applying the mixin. Note, not everywhere we need to apply it
    .crm-civirule-rule_label-block,
    .crm-civirule-trigger-block {
      @include block-form();

      // Table above will always have rows with bottom borders,
      // so we remove the top border to eliminate the border duplication.
      .crm-submit-buttons {
        border-top: 0;

    // Fix for the "Linked Trigger" container - it appears on both
    // "CiviRules Add Rule" and "CiviRules Update Rule" pages,
    // but on the "CiviRules Update Rule" it should not have vertical gaps
    #civirule-triggerBlock-table {
      margin-bottom: -#{$crm-standard-gap};
      margin-top: -#{$crm-standard-gap};
  ...

  &.page-civicrm-civirule-form-action {
    // There is a switch between `.groups-single` and `.groups-multiple`
    // and `.groups-single` is not a last child, but doesn't need that margin
    .groups-single {
      margin-bottom: 0;
  
  ...

  // "CiviRules Add Action" page
  // we style the "time" inputs
  &.page-civicrm-civirule-form-rule-action {
    #XWeekDay_time_hour {
      text-align: right;

    #XWeekDay_time_hour,
    #XWeekDay_time_minute {
      width: 20px;
  ...
}
```

However, some selectors cannot even be isolated for extension because `page-civicrm-civirule` class does not exist on these pages.

```css
// "Find CiviRules Rules" page
// This is the only reliable selector for this page because it is built as a
// Custom Search page and it does not contain any Rules extension specific classes
#civirule_helptext_dialog-block + .crm-form-block {
  @include block-form();

  padding-bottom: 0 !important;
  padding-top: 0 !important;

// Help icon in "Description" column on "Find CiviRules Rules" results page
#civirule_help_text_icon {
  margin-left: $crm-standard-gap / 4;
```

### ⚠️New class `.crm-form-block-container`

There is a problem with form inside widely used block which looks like this:

image

As you may notice, there are no paddings (+ some other minor issues). We could apply paddings globally, but sometimes we do not need them:

image

We could of course apply negative margins to the inner table, but not every table is needed to be like this, sometimes there are tables which are actually forms and identifying them is not an easy thing, hence the inconsistency in Civi markup.

So, to apply paddings and other fixes, we need to identify that block and apply styles. However, sometimes identifying such block are either impossible or unacceptable for Shoreditch, for example, if these blocks are a part of custom extension (we cannot identify a div by a class like `.this-is-my-cool-extension` in Shoreditch, because Shoreditch is a theme, not an extension styler).

In this particular situation, we **need** to style a 3rd party extension block. That's why we provide this class that can be applied to a container of the form and, thus, style it.

```pug
.crm-form-block-container
  h3 Some title
  .crm-form-block.crm-block
    //- some form
```

```css
.crm-form-block-container {
  .crm-form-block {
    @include block-form();
```

Once you apply it, it will look like this:

image

### ⚠️Global sorting icons table styling

The double-caret icon should not be offset, only single-caret icons should:

```css
a.sorting,
a.sorting_asc,
a.sorting_desc {
  ...

  &::after {
    ...
    // removed top -3px from here

  // only apply to single carets!
  &:not(.sorting) {
    &::after {
      top: 3px; 
```

## Testing

Manually tested.

### DIFF
--- a/scss/civicrm/common/_forms.scss
+++ b/scss/civicrm/common/_forms.scss
@@ -328,3 +328,11 @@ input.error {
     border-color: $input-border-focus;
   }
 }
+
+// Add this class to style a form inside a container
+// like you see at /civicrm/civirule/form/rule?reset=1&action=add
+.crm-form-block-container {
+  .crm-form-block {
+    @include block-form();
+  }
+}
--- a/scss/civicrm/_mixins.scss
+++ b/scss/civicrm/_mixins.scss
@@ -2,6 +2,7 @@
 
 @import 'mixins/accordion';
 @import 'mixins/advanced-multiselect';
+@import 'mixins/block-form';
 @import 'mixins/border-box';
 @import 'mixins/calc';
 @import 'mixins/civicrm-form-control';
--- a/scss/civicrm/_mixins.scss
+++ b/scss/civicrm/_mixins.scss
@@ -21,5 +21,6 @@
 @import 'mixins/section-title';
 @import 'mixins/table-block';
 @import 'mixins/table-header';
+@import 'mixins/table-selector';
 @import 'mixins/transform';
 @import 'mixins/translate';
--- a/scss/civicrm/_variables.scss
+++ b/scss/civicrm/_variables.scss
@@ -1,5 +1,7 @@
 /* stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type, selector-max-id */
 
+$crm-standard-gap: 20px;
+
 $contact-avatar: 90px;
 $crm-control-height: 30px;
 
--- a/scss/civicrm/common/_base.scss
+++ b/scss/civicrm/common/_base.scss
@@ -227,6 +227,19 @@
       width: 50%;
     }
   }
+
+  /**
+   * @NOTE the table that goes straight after <h3> does not
+   * need a shadow because in that case the shadow falls onto the <h3>.
+   * The shadow instead is applied
+   * to the *container* of the <h3> and <table>.dataTables_wrapper
+   *
+   * @TODO remove all the .dataTables_wrapper { box-shadow: none; }
+   * from "per-page" styles and other files
+   */
+  h3 + .dataTables_wrapper {
+    box-shadow: none;
+  }
 }
 
 .page-civicrm {
--- a/scss/civicrm/common/_base.scss
+++ b/scss/civicrm/common/_base.scss
@@ -302,7 +302,12 @@ a.sorting_desc {
   &::after {
     position: absolute;
     right: 0;
-    top: 3px;
+  }
+
+  &:not(.sorting) {
+    &::after {
+      top: 3px;
+    }
   }
 }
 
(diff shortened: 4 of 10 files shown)


scss/civicrm/common/_forms.scss
```diff
@@ -328,3 +328,11 @@ input.error {
     border-color: $input-border-focus;
   }
 }
+
+// Add this class to style a form inside a container
+// like you see at /civicrm/civirule/form/rule?reset=1&action=add
+.crm-form-block-container {
+  .crm-form-block {
+    @include block-form();
```
AkA84: this looks like a global change, anything reported by backstop?

igorpavlov: `.crm-form-block-container` is not used anywhere **yet**, sorry, should have described a bit in more details. Basically, @tunbola will need this class to apply to his changes in a **custom** extension. I **cannot** do in Shoreditch smth like

```css
.tunbolas-custom-extension-wrapper {
  .crm-form-block {
    @include block-form();
```
because it is not related to vanilla civi. So I added a new class that he + other devs can use to style the form inside the accordion.

PR #381, opened 2019-03-21
SHOR-102: Fix leaking font-family

### LINKED ISSUE #136: custom font in Drupal overridden
The last line of the css: {font-family:'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif !important}

is used with the !important on the very general body element so it is overriding the font used by Drupal on all pages. It does not seem needed as this font-family is already assigned to all Civicrm pages (on body class .page-civicrm).

### LINKED ISSUE #234: Please don't apply font-family to the body element
Currently, Shoreditch [applies its sans-serif font stack to the body element](https://github.com/civicrm/org.civicrm.shoreditch/blob/master/scss/civicrm/common/_base.scss#L8). This seems, firstly, to be a divergence from the aim to "contain" Shoreditch's CSS declarations to CiviCRM markup. Secondly, on WordPress, this causes problems both in wp-admin (because it overrides the declarations set by WordPress itself) and on the front-end (because it overrides the theme's default font stack). 

In wp-admin, my CiviCRM Admin Utilities plugin resets the font stack to that which WordPress itself provides, but on the front-end it cannot know what the theme declares since each may be different.

I am aware that the README says that Shoreditch is Drupal-only but there are lots of installs which already use it on WordPress, so removing this declaration seems like a simple fix to prevent unnecessary incompatibility.

AkA84:
Closes #136, Closes #234 

## Problem
The theme was leaking the Open Sans font family outside the CiviCRM boundaries (`.crm-container`, `#civicrm-menu`, `#root-menu-div`), causing the font family to be forcefully applied also to non-civi pages

### Example on Drupal
drupal-before

### Example on Wordpress
wordpress-before

The issue was caused by the `font-family` property being assigned directly to the `body`
```scss
html {
  body {
    @include open-sans-font-family();
    // ...
  }
}
```
This was originally done in order to apply the same font family to the `#branding` element of the Drupal's Seven theme (the only officially supported cms+theme by the theme at the moment), which is not an element "owned" by Civi but that nonetheless for consistency's sake needed to use the same font

boundaries

Additionally, the `open-sans-font-family()` SASS mixin was too aggressive, using `!important` when declaring the `font-family` value

```scss
@mixin open-sans-font-family() {
  font-family: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif !important;
}
```

## Solution
In order to fix the leak, the `font-family` declaration had been isolated within `.crm-container` (for CiviCRM) and `#branding` (for the Seven's `#branding` element)

Additionally, the `!important` had been removed as it wasn't really necessary

### Example on Drupal
drupal-after

### Example on Wordpress
wordpress-after

```scss
// common/_base.scss

.crm-container {
}
```
```scss
// cms/_drupal.scss

.page-civicrm {
  #branding {
  }
}
```

### Removed `open-sans-font-family()` mixin
The `open-sans-font-family()` mixin was removed as it wasn't providing any tangible value in terms of reusability/encapsulation (it was simply wrapping a `font-family` declaration and nothing else).

```diff
.crm-container {
-  @include open-sans-font-family();
+  font-family: $font-family-base;
}
```

### Consolidated drupal-related styles under a `drupal.scss` partial
There were several drupal-related styles (= styles targeting elements owned by drupal, or by the Seven theme) scattered across different SASS partials, sometimes even duplicated or diverging

_(Following are just the selectors, it's not important to show the code in this case)_

```scss
// common/_alerts.scss

.page-civicrm div#console {}
```
```scss
// common/_base.scss

html {
  body {
    &:not(.toolbar) {}
    .toolbar-menu {}
  }
}

.page-civicrm {
  #branding {
    // several other selectors nested here
  }
}
```

```scss
// common/_top.scss

.page-civicrm {
  #toolbar {}
  #branding {}
  h2 {}
  #console {}
  div#printer-friendly {}
}
```
_(The only exception would be `div#printer-friendly`, given that that element does belong to CiviCRM, but given that Shoreditch changes its position in relation to the Seven theme, its can still be considered a drupal-related style)_

All the styles are now consolidated under a `cms/_drupal.scss` partial, as a way to separate them from the rest of the civi-specific styles

## Comments
The fact that Shoreditch concerns itself with styling things outside of Civi boundaries (even after this PR) is obviously not ideal.

For this reason we added to the [pipeline](https://lab.civicrm.org/extensions/shoreditch/issues/2) the creation of a dedicated drupal theme for Shoreditch, which will allow the decoupling of the drupal-related styles.

That's also the reason for the `drupal.scss` partial, so that it will be easier to migrate the styles to the new theme


scss/civicrm/cms/_drupal.scss
```diff
@@ -0,0 +1,109 @@
...
+    @extend %title;
+  }
+
+  #branding {
+    background: $crm-page-topbar-bg !important;
+    color: $crm-white !important;
+    font-family: $font-family-base;
+    line-height: 20px;
+    overflow: visible !important;
+    padding: 20px 0 0;
```
igorpavlov: this is probably '$crm-standard-gap' var

scss/civicrm/cms/_drupal.scss
```diff
@@ -0,0 +1,109 @@
...
+
+  h2 {
+    @extend %title;
+  }
+
+  #branding {
+    background: $crm-page-topbar-bg !important;
+    color: $crm-white !important;
+    font-family: $font-family-base;
+    line-height: 20px;
```
igorpavlov: this is probably '$crm-standard-gap' var

scss/civicrm/cms/_drupal.scss
```diff
@@ -0,0 +1,109 @@
...
+
+    h1.page-title { /* stylelint-disable-line selector-no-qualifying-type */
+      @extend %title;
+
+      background: $crm-white;
+      box-sizing: border-box;
+      color: $gray-darker;
+      display: block;
+      line-height: 50px !important;
+      margin: -20px 0 20px;
```
igorpavlov: same as above.
also, there are other places in this PR where you use 20, 40 etc, if it relates to the "grid" may be we should use the '$crm-standard-gap' as well?

PR #386, opened 2019-04-11
SHOR-136: Support KAM menu

AkA84:
Closes https://github.com/civicrm/org.civicrm.shoreditch/pull/364

## Overview
The [uk.squiffle.kam](https://github.com/aydun/uk.squiffle.kam) extension, which implements an improved and responsive main menu, is now part of core as of CiviCRM v5.12.

Shoreditch's CSS is targeting the old menu still, so it needs to be updated

## Before
The menu looks broken in places, due to the theme's styles targeting the old menu (with which the new menu shares some markup and CSS classes)

with-shoreditch-issue-1

with-shoreditch-issue-2

## After
The menu is not broken anymore, and the theme applies its color palette, fonts, and horizontal rhythm to it

#### Desktop view, with sub menu and search results open
with-shoreditch-fixed

#### Search options
with-shoreditch-search-options-fixed

#### Mobile view
with-shoreditch-mobile-fixed
  
### ⚠️ Menu Height
The height of the menu is kept as-is, instead of being increased as it was done for the old menu

menu-height-old

There are several reasons for us to avoid changing the height again, which can be shortly summarized by [this comment](https://github.com/civicrm/org.civicrm.shoreditch/pull/364#issuecomment-472521742)

### CSS / JS changes
All of the old CSS and JS code have been deleted, and then the minimum amount of CSS had been applied to match the Shoreditch's look-and-feel

The KAM menu has a couple of media query breakpoints ([desktop](https://github.com/civicrm/civicrm-core/blob/master/css/crm-menubar.css#L185), [mobile/tablet](https://github.com/civicrm/civicrm-core/blob/master/css/crm-menubar.css#L248)), which we are mirroring in order to correctly apply our customizations

```scss
// civicrm/common/_main-menu.scss

#civicrm-menu.sm {
  @media (max-width: $screen-sm-min) { }
  @media (min-width: $screen-sm-min)
}
```
unfortunately civi core does not use SASS and as such they don't expose their breakpoints in variables we can reuse in our theme, but at least their breakpoints match the ones Bootstrap uses

### Overriding `v5.12.1` menubar color settings
In https://github.com/civicrm/civicrm-core/pull/13996/ (released as part of `v5.12.1`) a new "Menubar color" setting under _admin/setting/preferences/display_ had been added, which allows a civi admin to personalize the color of the main menu

This of course doesn't work very well with Shoreditch, which is pretty opinionated in terms of color scheme. This is an example of the a civi site with shoreditch enabled, and with the menu color set to green:

menu-settings

The solution is to have the theme always override (for desktop-like widths) the user's chosen menu color (after all, the theme is there to enforce a certain look-and-feel)

```scss
@media (min-width: $screen-sm-min) {
  li {
    a {
      background: $crm-white;
    }
  }

  ul {
    background: $crm-white;
  }
}
```

### Upgrading, hinting, and enforcing node `v10.15.3`
The recommended node version had been bumped from `v8.9.4` (old LTS) to `v10.15.3` (new LTS). No unintended changes in the .css files had been detected after upgrading and reinstalling the npm modules

Additional, a `.nvmrc` file with the version hint had been added,  see https://github.com/creationix/nvm#nvmrc for additional details

Finally, the new recommended version is now being enforced with changes to the `package.json` file
```json
{
  "engines": {
    "node": "~10.15.3"
  }
}
```
and by adding a `.npmrc` file
```
engine-strict=true
```
see https://stackoverflow.com/questions/29349684/how-can-i-specify-the-required-node-js-version-in-package-json/50798589#50798589 for additional details

## Additional comments
The minimum compatible version for the extension is now `5.8` (as it was the minimum version that the KAM extension would work on)
```xml
<compatibility>
  <ver>5.8</ver>
</compatibility>
```
the README file had been updated to reflect the change, with `v5.12` now the recommended version


igorpavlov:
@AkA84 

**Search options**
The gap between radios and text is imho a bit too narrow. Would we like to increase it a bit and confirm how it should look like?

**Mobile view**
Imho, doesn't really look good. The center alignment is smth that doesn't just work. **Imho**. Wouldn't it be nicer if we have a fixed width container (same width as the search input) and align left? Again, needs to be confirmed.

package.json
```diff
@@ -12,6 +12,9 @@
   "scripts": {
     "postinstall": "npx gulp copy && npx gulp sass"
   },
+  "engines": {
+    "node": "~10.15.3"
```
igorpavlov: Please used fixed version `10.15.3` (no tilde)

package.json
```diff
@@ -21,18 +24,22 @@
...
     "gulp-postcss": "^7.0.1",
     "gulp-replace": "^1.0.0",
-    "gulp-sass": "^4.0.1",
+    "gulp-sass": "^4.0.2",
     "gulp-sass-glob": "^1.0.9",
     "gulp-strip-css-comments": "^2.0.0",
     "gulp-transform-selectors": "0.0.2",
     "husky": "^1.0.0",
     "lint-staged": "^6.1.1",
+    "npm-force-resolutions": "0.0.2",
```
igorpavlov: Why do you use fixed version here and not in other places? I think we should have fixed versions everywhere. It is pretty common when a minor version update is not properly working.

scss/civicrm/common/_main-menu.scss
```diff
@@ -0,0 +1,109 @@
+/* stylelint-disable selector-max-id, max-nesting-depth, selector-no-qualifying-type, selector-max-compound-selectors */
+
+#civicrm-menu.sm {
+  background: $crm-white;
+  font-family: $font-family-base;
+
+  li {
+    line-height: 1.3;
```
igorpavlov: Please extract this to a separate var and put it on top of this file.

scss/civicrm/common/_main-menu.scss
```diff
@@ -0,0 +1,109 @@
...
+    } 
+  }
+
+  #crm-qsearch-input {
+    border-radius: $border-radius-base;
+    font-family: $font-family-base;
+  }
+
+  .crm-i {
+    margin-right: 6px;
```
igorpavlov: Same, please create a var.

AkA84:
@colemanw could you test this if you have time (I have tested it locally against 5.12 and 5.11 + kam extension)? Thanks!

colemanw:
This seems good for versions 5.8 - 5.11, however due to a number of complaints about the new color of the KAM menu in core, in 5.12.1 we are adding a new configurable menubar color feature: https://github.com/civicrm/civicrm-core/pull/13996

Screenshot

So I think the preferred way for Shoreditch to turn the menu white would be to add this to the installer:

    \Civi::settings()->set('menubar_color', '#ffffff');

This will turn the menu white while still respecting the setting in case users wish to change it (hardcoding the color in your css file here will potentially frustrate users by ignoring their chosen color).

jamienovick:
@colemanw Can you share what the complaints were? My overall sense is that KAM does look a bit ugly and we have some work to do to get it to look ok with Shoreditch now... given that Shoreditch is all about the aesthetic...

AkA84:
@colemanw thanks for taking a look, about

> So I think the preferred way for Shoreditch to turn the menu white would be to add this to the installer:
> ```
>   \Civi::settings()->set('menubar_color', '#ffffff');
> ```

but that would only work if the user has v5.12.1, wouldn't it? If that's the case, then for users on `v5.12.0` (or lower) it would not have any effect, so we would still have to set the color via CSS

package.json
AkA84: I trusted [this comment](https://stackoverflow.com/questions/29349684/how-can-i-specify-the-required-node-js-version-in-package-json/50798589#comment97673889_50798589) but now I've checked and it still works fine even if we don't specify a range. Will fix

package.json
AkA84: > Why do you use fixed version here and not in other places?

This line has been generated by 
```
npm i --save-dev npm-force-resolutions
```
(same for `gulp-transform-selectors`, which doesn't have a range)

>  I think we should have fixed versions everywhere

No need, that's what the `package-lock.json` file is for

colemanw:
@jamienovick you're the first to call it _ugly_ :scream: 

The complaints were about consistency - users who are accustomed to a black menu or are relying on documentation specifying "go to the black menu" would be confused.

eileenmcnaughton:
Lol - I think there might be some taste involved - FWIW I think the new menu in core looks great

colemanw:
@jamienovick with the release of 5.12.1 we now have configurable menubar colors. I'd be happy to backport that feature to the KAM extension and then Shoreditch can always rely on that setting being there.

AkA84:
@colemanw that would be very helpful, so that for `v5.12.0` and lower at least we can point to the latest KAM version as a dependency.

What would be the ETA?

AkA84:
@colemanw would the following work also on `v5.12.0` (so, when using the KAM extension)?

```php
function shoreditch_civicrm_enable() {
  _shoreditch_civix_civicrm_enable();

  \Civi::settings()->set('menubar_color', '#ffffff');
}

function shoreditch_civicrm_disable() {
  _shoreditch_civix_civicrm_disable();

  $default_color = \Civi::settings()->getDefault('menubar_color');
  \Civi::settings()->set('menubar_color', $default_color);
}
```

More specifically i worry about being able to get the original menu color both on `v5.12.0` and on `v5.12.1`, which I don't know if it's something that will be possible given that on the former that setting doesn't exist in `Core.setting.php` yet

AkA84:
@colemanw actually after experimenting with it a bit, I think we really don't need the above in the end.

Given that the theme needs to enforce its own color scheme anyway (which applies to [items bg color](https://github.com/civicrm/civicrm-core/pull/13996/files#diff-ff63186dcddd8213d6b992efd57d70a4R856), [highlight color](https://github.com/civicrm/civicrm-core/pull/13996/files#diff-ff63186dcddd8213d6b992efd57d70a4R857), etc), we can simply do that with our own css, without having to change the menu color setting

AkA84:
@colemanw could you test the PR again? I've tested it against
* `v5.11.0` (new menu provided by KAM extension)
* `v5.12.0` (new menu in core)
* `v5.12.3` (new menu in core and menu color setting)

and it all looks consistent to me

Thanks!

colemanw:
@AkA84 and @jamienovick I've taken a first stab at making the height of the menubar adjustible.
CiviCRM does not use SASS but we do have a concept of "preprocessing" certain files and in the process we can do variable substitution. But the mechanism by which an extension like shoreditch would be able to *alter* the value of the variables is something we have yet to figure out.
https://github.com/civicrm/civicrm-core/pull/14122

AkA84:
> But the mechanism by which an extension like shoreditch would be able to alter the value of the variables is something we have yet to figure out.

@colemanw wouldn't what you suggested for the menubar color work also in this case?

> So I think the preferred way for Shoreditch to turn the menu white would be to add this to the installer:
> ```
>   \Civi::settings()->set('menubar_color', '#ffffff');
> ```

---

> so shoreditch could change the default but still allow overrides

@eileenmcnaughton I'm not sure though that that's what we want, as Shoreditch as a theme needs to be opinionated and enforce a certain look-and-feel that the user should not be able to override.

For example in this PR I've overridden the menu color (and all colors that depend on it, like highlight color, text color, items color, etc) to match, and enforce, the original theme's color scheme. I think otherwise that it wouldn't make much sense having Shoreditch applying a certain color scheme across all pages if then the user can set the main menu to be pink!

eileenmcnaughton:
@AkA84 right - although if a user decides to change something - e.g change the menu height to reclaim some real estate that's not your problem really. You can certainly block it on sites you actually manage.

AkA84:
> change the menu height to reclaim some real estate that's not your problem really

@eileenmcnaughton that in theory could be up for debate, as if we feel that a specific height for the menu is important for the "identity" of the theme, then it would be a problem if the user has the freedom to go and change it

anyway this is more of a product-wise rather than technical-wise discussion, something that we will have to figure out internally, but thanks a lot for your comments!

eileenmcnaughton:
@AkA84 yep for sure - at the end of the day it's a preference thing for me - I just use my laptop screen & I find the real estate issues with shoreditch cause a bit of scrolling. But our users mostly have bigger screens,

We already mess up your colour scheme beyond what you would want to know about with our drupal theme & our lurid environment indicator :-)

AkA84:
> We already mess up your colour scheme beyond what you would want to know about with our drupal theme & our lurid environment indicator :-)

@eileenmcnaughton actually that sounds interesting, anything i can take a look at?

eileenmcnaughton:
@AkA84 we just use the drupal environment indicator - which adds a menu bar at the top - we chose colours at random for prod & staging. Staging is a 'tasteful' green. Prod was hot pink until it accidentally got changed to lilac recently.I think the theme is tivy  - it gives us more real estate than some other themes we've tried,

AkA84:
got it, thanks @eileenmcnaughton 

@colemanw did you have a chance to test the PR again? See https://github.com/civicrm/org.civicrm.shoreditch/pull/386#issuecomment-485814910

colemanw:
FYI I've added https://github.com/civicrm/civicrm-core/pull/14135 which will make the menubar height adjustable via hook.

FWIW I still disagree with the principle of hardcoding the menubar color. Many people have legitimate reasons for wanting to change it for example as an environment indicator on their test/staging sites. But this is your theme so it's not my call.

aydun:
> FWIW I still disagree with the principle of hardcoding the menubar color. Many people have legitimate reasons for wanting to change it for example as an environment indicator on their test/staging sites. But this is your theme so it's not my call.

+1  Shoreditch makes lots of changes  - layouts, heights, colours etc - but we don't necessarily want exactly your choices.  If I'm understanding @aka84's comments correctly the vision is for all sites running Shoreditch to look identical ...  if so, that severely limits Shoreditch's usefulness.  Remember Henry Ford?  "Any color so long as it's black"

jusfreeman:
Just a FYI. Sites upgrading to CiviCRM 5.13 now and using Shoreditch will be seeing this bug, https://github.com/civicrm/org.civicrm.shoreditch/issues/391

colemanw:
@jamienovick this incompatibility problem is affecting more and more people, so whether you want to consider my suggestion of allowing the color to be configured or not, I think we need to finish this and merge it so a 5.12-compatible version of Shoreditch can be released.

jusfreeman:
@colemanw @jamienovick any chance of getting this one resolved sometime soon?

Shwele:
@colemanw @jamienovick  any update on this topic? I've been following it for weeks now and we still didn't do update of CiviCRM past 5.11 because of this visual bug.

Hopefully its on their way, just bumping topic.

eileenmcnaughton:
@AkA84 what is blocking you merging this?

eileenmcnaughton:
Thanks @AkA84

jensschuppe:
Maybe it's a bit too late here, but ... my 5 cents to menubar color:

If there's a setting to change the menubar color, this is part of CiviCRM Core _functionality_ and thus not part of the _theme_ anymore. Shoreditch should respect this setting (with its own default, maybe) or give the user the choice to configure whether Shoreditch overrides the menubar color or not.

Bluntly overriding it without notice causes confusion for users.

colemanw:
I've made a core patch that would facilitate changing the menubar colors from shoreditch in a way that respects the setting. https://github.com/civicrm/civicrm-core/pull/14944

FYI another core patch I recently did would allow Shoreditch to adjust the menubar height via hook.

PR #389, opened 2019-04-18
Update _ui-select.scss

sleewok:
The current value blocks the user interface for adding fields, etc in Drupal Views. Setting z-index: 99 does not block it.


deb1990:
@sleewok 
Thanks for the contribution. But could you update the PR following the PR template https://github.com/civicrm/org.civicrm.shoreditch/blob/master/.github/PULL_REQUEST_TEMPLATE.md.

The Before/After screenshots will be specifically necessary to determine the outcome of the PR.

Issue #374, opened 2019-03-08
Contact images broken by contact-summary.js when using Drupal 'seven' theme.


mikantchap:
CiviCRM 5.11.0
Shoreditch 0.1-alpha28 (required for Mosaico)
Using 'Seven' as the CiviCRM admin theme.
Contact images are briefly displayed on eg civicrm/contact/view?reset=1&cid=22437
contact-summary.js is still loaded from
sites/default/files/civicrm/extensions/org.civicrm.shoreditch-0.1-alpha28/js/contact-summary.js?r=zQ7oF

It appears to move the contact image from 
<div id="crm-contact-thumbnail">
to further up the page where it is now invisible (probably because of CSS).

If the js is blocked the image shows in the correct postion.

Upperholme:
I'm seeing this issue with 5.10.x (and earlier versions) using Adminimal as the CiviCRM theme. The contact-summary.js file is getting loaded even though I'm only using Shoreditch for Mosaico.

samuelsov:
I confirm the problem. We are using adminimal as the CiviCRM theme and Shoreditch for Mosaico only.

Caused by https://github.com/civicrm/org.civicrm.shoreditch/issues/215

Upperholme:
"If the js is blocked the image shows in the correct position." How might I achieve this?

mikantchap:
In Chrome, inspect => network tab.
Filter so that only JS shows in list.
Reload page => now have list of JS resources
Right click contact-summary.js and use 'Block request URL'
Reload page (but contact-summary.js is blocked from loading)

jmcclelland:
I just pushed a patch so that the contact summary javascript is only pulled in if the site is configured to use shoreditch throughout civicrm.

deepak-srivastava:
Similar problem with us. Images don't appear at all. 

If we block the js/contact-summary.js file from loading OR comment out the code in the js file, image gets displayed fine.

![image](https://user-images.githubusercontent.com/33451769/92463893-4b740480-f1c4-11ea-9f9f-b0ee2823ac44.png)

![image](https://user-images.githubusercontent.com/33451769/92463645-fcc66a80-f1c3-11ea-8fc0-865037fa69b6.png)

jitendrapurohit:
Can replicate the same error on 

Drupal 9.2.6
CiviCRM 5.41.1
shoreditch 1.0.0-beta.9

[...]

PR #396, opened 2019-07-08
don't try to move image if we are not using shoreditch

jmcclelland:
Fix for #374 if you are only using shoreditch for mosaico.


mlutfy:
I think this isn't necessary anymore, with the introduction of the "backend theme" setting.

Since shoreditch beta1, it's not possible anymore to enable shoreditch, but not use the custom CSS. Either you set the backend theme to shoreditch, or it doesn't load. (i.e. previously we could enable shoreditch, without the CSS, to that Mosaico was usable)

jmcclelland:
I agree, no longer relevant to us since now everyone is using shoreditch in Powerbase.

PR #398, opened 2019-07-25
SHOR-144: Remove drupal partial

AkA84:
_(Related to https://github.com/compucorp/shoreditch-companion-d7-theme/pull/1)_

## Overview
With the new drupal companion theme (see PR linked above), shoreditch doesn't need to style elements outside of the civi container(s) like page titles, breadcrumbs, etc anymore.

This allows the theme to be decoupled by the CMS and to just focus on styling CiviCRM. In fact now the "Seven" Drupal theme is not mandatory anymore, while the new companion theme is recommended but not obligatory to use.

Additionally, the decoupling solved some style leaking (see Before/After sections for an example)

## Before

leak-before

_(Public page of an event)_

leak-admin-before

_(Admin page, with a theme other than "Seven" as "CiviCRM Administration theme")_

The `#branding` element looks wrong because shoreditch applies a style to it despite it belonging to Drupal, not CiviCRM

Shoreditch going "out of bounds" went unnoticed on the civi admin pages as it was mandatory requirement to use the "Seven" theme as the "CiviCRM administration theme"

seven-before

_(Admin page, with "Seven"  as "CiviCRM Administration theme". Notice that although the page looks fine shoreditch still styles the `#branding` element)_

but one just needed to go to a public page (like an event) or to change the "CiviCRM Administration theme" to some other theme to surface the issue

## After
leak-after

_(Public page of an event)_

lead-admin-after

_(Admin page, with a theme other than "Seven" as "CiviCRM Administration theme")_

## Technical details
Removed the whole [scss/civicrm/cms/_drupal.scss](https://github.com/civicrm/org.civicrm.shoreditch/pull/398/files#diff-6827d3f950b1af76159e4c331dee5659) partial as its content now belongs to the drupal companion theme

Removed references to `#page` and `h1.page-title` in the [scss/civicrm/common/_backgrounds.scss](https://github.com/civicrm/org.civicrm.shoreditch/pull/398/files#diff-57e18a98a6f84590d0c08b242f50c45d) partial as those are drupal's elements, not civi's


README.md
```diff
@@ -10,13 +10,13 @@ Please note that this extension is currrently in **alpha stage** and under **act
 * CiviCRM `v5.8`+, `v5.12`+ recommended
 * [uk.squiffle.kam](https://github.com/aydun/uk.squiffle.kam) if the CiviCRM version is lower than `v5.12`
 * "CiviCRM Theme" module enabled
-* "Seven" theme set as CiviCRM Administration theme
+* The [CiviAdmin Companion Drupal theme](https://github.com/compucorp/shoreditch-companion-d7-theme) is optional but  recommended
```
deb1990: Looks like there is 2 spaced between `but` and `recommended`.

README.md
```diff
@@ -10,13 +10,13 @@ Please note that this extension is currrently in **alpha stage** and under **act
...
 
 ### For development
 * NodeJS `v10.15.3`
 
 ## Supported CMSs
-At the moment the theme is being developed to work only in Drupal 7 (with default theme set to "Seven"). WordPress and Joomla are not currently supported.
+At the moment the theme is being developed to work only in Drupal 7. WordPress and Joomla are not currently supported.
```
deb1990: Is this text necessary anymore? Because after decoupling, shoreditch should not care about what CMS is supports.

README.md
```diff
@@ -26,12 +26,12 @@ cd sites/default/files/civicrm/ext
...
 git clone https://github.com/civicrm/org.civicrm.shoreditch
 cv en org.civicrm.shoreditch
-
-# Enable the "CiviCRM Theme" module and set "Seven" as the Civi admin theme
-drush en -y civicrmtheme
-drush vset civicrmtheme_theme_admin seven
-drush cc civicrm
 ```
+
+### Companion theme
```
deb1990: We might change the text in the section to something like
```
### Companion theme for other CMS's
Drupal 7 - [CiviAdmin Companion Drupal theme](https://github.com/compucorp/shoreditch-companion-d7-theme)

Other CMS's are not supported yet.
```

This way the above block is not necessary anymore, as this serves both. And in future we can add other CMS's theme link when they are ready.

Issue #387, opened 2019-04-11
Changes styling of public-facing event info pages


Upperholme:
I've recently installed Shoreditch on a Drupal 7/Civi site and noticed that the styling of my public-facing event info pages is being modified. I thought Shoreditch was for admin use only?

AkA84:
hi @Upperholme , what version are you using?

Upperholme:
0.1-alpha31

AkA84:
@Upperholme thanks, could you attach some screenshots so we can see what part of the styling is being affected? any steps to reproduce the issue would be useful as well

Upperholme:
First screenshot, with white background, shows an event info page with Shoreditch turned off. Image with non-white background shows the same page with Shoreditch on. Also seeing impacts elsewhere on the site on public-facing pages, notably that enforced padding at the top of the viewport is appearing on my front page at least. This is for a non-logged in user. 

![Screenshot 2019-04-11 at 15 42 14](https://user-images.githubusercontent.com/5860951/55966535-7fb26100-5c70-11e9-9284-92d761b23fc3.png)
![Screenshot 2019-04-11 at 15 40 58](https://user-images.githubusercontent.com/5860951/55966536-7fb26100-5c70-11e9-95f2-c55f5b31d82e.png)

Upperholme:
What I'm unclear about is why Shoreditch is having any impact at all on public-facing pages. Why is the stylesheet being loaded on these pages?

jamienovick:
@AkA84 @Upperholme  I think this may be a combination of issues actually manifesting itself.

1. I believe that CiviCRM does not differentiate between public and private pages in its CSS. In fact all CiviCRM CSS is loaded on public CiviCRM pages in general anyway. They just normally don't have a lot of styling so its less obvious. See here for discussion on this issue: https://lab.civicrm.org/dev/core/issues/300 

2. This is then made worse as Shoreditch more aggressively styles the page and sometimes outside of the main CiviCRM divs. The buttons and and page margin/padding at the top is then affected as well on any page where CiviCRM is bootstrapped.

I think the long term solution to this is to solve 1 above but we have been short of time/budget to do so. I'm not sure if @Upperholme this is something you could help look at perhaps and help us progress?

Upperholme:
[...]

PR #399, opened 2019-07-29
SHOR-146: Use theme system, set shoreditch as backend theme only

AkA84:
## Overview
The soon-to-be-released Civi v5.16 introduced a [theme system](https://docs.civicrm.org/dev/en/latest/framework/theme/), which allows among other things the ability to decide whether to inject the theme in public pages or not, which in turns would solve the issue of style leaks in public pages (see https://github.com/civicrm/org.civicrm.shoreditch/issues/387)

For this reason we are migrating `org.civicrm.shoreditch` from a simple extension to a proper theme

Additional infos: https://lab.civicrm.org/dev/core/issues/378

## Before
public-before

The shoreditch theme is being injected into public pages, causing a style leak

## After
public-after

The shoreditch theme gets defined as a backend theme only, resulting in its stylesheets not being injected into public pages anymore

### Code update
Both `shoreditch.php` and `shoreditch.civix.php` had been changed to use the most up-to-date code that the `civix generate:module` command generates. This was necessary in order to be able to use the new theme system

### Add theme metadata
A `shoreditch.theme.php` file had been added, which contains the theme metadata. The only thing worth mentioning is the `url_callback` property (see below)
```php
return array (
  'name' => 'shoreditch',
  'title' => 'Shoreditch',
  'url_callback' => '_shoreditch_css_url',
  // ...
);
```

### Url callback
A `_shoreditch_css_url` callback (see [here](https://docs.civicrm.org/dev/en/latest/framework/theme/#url_callback)) is being defined, for two reasons:
1. We need to specify that the theme it doesn't want to override the original `civicrm.css`, but rather build on top of it
2. We need to point civi to use the theme's `custom-civicrm.css` stylesheet, otherwise it would look for a `civicrm.css` file that doesn't exist

```php
// Simplified snippet

function _shoreditch_css_url($themes, $themeKey, $cssExt, $cssFile) {
  $urls = array_merge(
    /* 1 */ Civi::service('themes')->resolveUrls('greenwich', $cssExt, $cssFile),
    /* 2 */ \Civi\Core\Themes\Resolvers::simple($themes, $themeKey, $cssExt, 'css/custom-civicrm.css')
  );

  return $urls;
}
```

### Upgrader
An upgraded was added to automatically perform two tasks
1. Remove shoreditch's `custom-civicrm.css` file as the value of the "Custom CSS URL" field
    ```php
    $customCSSUrl = Civi::settings()->get("customCSSURL");
    
    if (strpos($customCSSUrl, 'org.civicrm.shoreditch') !== FALSE) {
      civicrm_api3('setting', 'create', [
        'customCSSURL' => ''
      ]);
    }
    ```
2. Set shoreditch as the backend theme only
    ```php
    civicrm_api3('setting', 'create', [
      'theme_frontend' => 'default',
      'theme_backend' => 'shoreditch'
    ]);
    ```


AkA84:
@totten I encountered 3 issues while adapting shoreditch to the new theme system, it'd be great if you could take a look, because I'm not sure if there is a problem with the system or if i'm getting something wrong, thanks!

### 1. Missing bootstrap.css
It seems that the `bootstrap.css` file is not being injected automatically. I tried also with a brand new theme and I still couldn't see the file on the page

**Before**
bootstrap-before

This is when the theme was "manually" injecting the file in the page head

```php
function shoreditch_civicrm_coreResourceList(&$items, $region) {
  if ($region == 'html-header') {
    CRM_Core_Resources::singleton()->addStyleFile('org.civicrm.shoreditch', 'css/bootstrap.css', -50, 'html-header');
```

**After**
bootstrap-after

Stylesheet is not injected in the page

### 2. Different file order
It seems that the `civicrm.css` stylesheet of a theme gets injected in a different position than the position the "Custom CSS URL" stylesheet was injected in, causing styling issues in shoreditch

**Before**
import-order-before

`custom-civicrm.css` as "Custom CSS URL" is being injected after `civicrm/css/searchForm.css`

**After**
import-order-after

`custom-civicrm.css` is now being injected before `civicrm/css/searchForm.css`, meaning that the latter overrides the rulesets of the former

### 3. Url callback code
In the [documentation](https://docs.civicrm.org/dev/en/latest/framework/theme/#url_callback), the code to instruct Civi to keep the original `civicrm.css` and just append the theme's afterwards is as follow:

```php
function _newyork_css_url($themes, $themeKey, $cssExt, $cssFile) {
  $urls = \Civi\Core\Themes\Resolvers::simple($themes, $themeKey, $cssExt, $cssFile);
  switch ("{$cssExt}/{$cssFile}") {
    case 'civicrm/css/civicrm.css':
      $urls = array_merge(
        Civi::service('themes')->resolveUrls('greenwich', $cssExt, $cssFile),
        $urls
      );
  }
  return $urls;
}
```
That unfortunately didn't work for me: while `$urls` was indeed made up of both greenwhich's and shoreditch's `civicrm.css` files, only the latter got injected into the page

I thought that maybe civi didn't like that both files had the same name, and in fact once i kept `custom-civicrm.css` as the shoreditch's file, then it worked
```php
function _shoreditch_css_url($themes, $themeKey, $cssExt, $cssFile) {
  switch ("{$cssExt}/{$cssFile}") {
    case 'civicrm/css/civicrm.css':
      $urls = array_merge(
        Civi::service('themes')->resolveUrls('greenwich', $cssExt, $cssFile),
        \Civi\Core\Themes\Resolvers::simple($themes, $themeKey, $cssExt, 'css/custom-civicrm.css')
      );
      break;
    default:
      $urls = \Civi\Core\Themes\Resolvers::simple($themes, $themeKey, $cssExt, $cssFile);
  }

  return $urls;
}
```

totten:
I've pushed up a few commits which the issues above. More details in commit notes.

igorpavlov:
Rebased with latest master

igorpavlov:
@totten I have reviewed the code, installed a local version of CiviCRM `5.16.0` and can confirm the issues pointed out by @AkA84 are addressed and resolved.

сс @jamienovick

jamienovick:
@totten do you think we will need to test those upgrading from previous versions for this? I think the upgrader probably does the job? @igorpavlov FYI.

totten:
@jamienovick  Well, since there is upgrade logic, it's a good idea to try out an upgrade (if it hasn't been tried otherwise).

:+1: on communicating clearly. 

One mitigating consideration... which doesn't help with folks pulling directly from Github, but... the `info.xml` update (a2f37920999d7f2b5576310bc0a367c9fb7bf30f) should help some other use-cases:

* On https://civicrm.org/extensions/shoreditch the "Releases" table mentions the minimum version.
* For anyone using pre-release feeds for semi-automatic download (`cv dl --dev` or `ext_repo_url`), the feed is filtered on compatibility. So `5.15` sites don't see releases that require `5.16`.

Related: #403

PR #401, opened 2019-08-06
SHOR-143: Style Mailing screens

igorpavlov:
# Overview

This PR styles the following screens in CiviCRM:
- Mailing Browse / Search - `/civicrm/mailing/browse/unscheduled?reset=1&scheduled=false`
- Mailing Reports - `/civicrm/mailing/report?mid=5&reset=1`

# Before

## Mailing Browse / Search

Before 1

## Mailing Reports

Before 2

# After

## Mailing Browse / Search

After 1

## Mailing Reports

After 2

# Technical Details

We did not amend any core markup to avoid unexpected changes in CiviCRM UI without Shoreditch. Instead, we applied **isolated** per-page CSS styles. The styles are trivial in most places but there is one specific to note explicitly:

```
#{civi-page('mailing-report')} {
  fieldset {
    legend {
      ... 
      // Otherwise it will actually be "detached" from the fieldset
      float: left;
```

## Why so many !importants?

The reason is that we can either same selectors as Civi or `!important`. In that particular case, when we are styling a page, `!important` has more sense because the code looks cleaner and we will not need to override these styles again, they are not global. Additionally, some styles coming from Civi actually have `!important` already.

# Tests

Manual. BackstopJS was **not** run because we have isolated the pages via CSS selectors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Please note that this extension is currrently in **alpha stage** and under **act
 
 ## Requirements
 * Drupal 7
-* CiviCRM `v5.8`+, `v5.12`+ recommended
+* CiviCRM `v5.16`+. In version `v5.16.0` CiviCRM introduces Theming (https://docs.civicrm.org/dev/en/latest/framework/theme) which makes Shoreditch no longer support the obsolete way of theme inclusion (since Shoreditch `v0.1-alpha34`) and drop the support for earlier CiviCRM versions.
 * [uk.squiffle.kam](https://github.com/aydun/uk.squiffle.kam) if the CiviCRM version is lower than `v5.12`
 * "CiviCRM Theme" module enabled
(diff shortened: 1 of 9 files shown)


jamienovick:
@igorpavlov  In the screenshot above there appears to be a small issue with the recipients block. Was that fixed in your PR and the screenshot is out of date or does it still need to be resolved?

igorpavlov:
@jamienovick yes, you are right, the margin was not committed, apologies. Here is how it looks now:

image

Also, please see the updated README: https://github.com/civicrm/org.civicrm.shoreditch/pull/401/files#diff-04c6e90faac2675aa89e2176d2eec7d8L9

swastikpareek:
Approved last 3 commits.

igorpavlov:
Rebased with master

PR #402, opened 2019-08-07
SHOR-139: Style CiviRules, GDPR and Pivot Reports extensions

igorpavlov:
# Overview

This PR styles the following extensions:
- CiviRules (https://civicrm.org/extensions/civirules)
- GDPR (https://civicrm.org/extensions/gdpr)
- Pivot Reports (https://civicrm.org/extensions/civicrm-pivot-reports)

# Before

## CiviRules
before 1
before 2

## GDPR
before 3
before 5
## Pivot Reports
before 4

# After

## CiviRules

after 1
after 2

## GDPR

after 3
after 5

## Pivot Reports

after 4

# Technical Details

All stylings are isolated (are applied to specific wrapping classes which guarantees the absence or regressions. There is nothing specific to mention - all stylings are trivial, except the shadows in CiviRules.

## CiviRules

We had to solve the consistency of shadows and the problem was that there are two consequent blocks (H3 header + body) that do not have a wrapper. This required to apply a pretty "hacky" way of applying shadows. The method is described here: https://codepen.io/igorpavlov/pen/Qeaezx.

```css
.page-civicrm-civirule {
  // These z-indexes are required because of how CiviCRM page is built
  position: relative;
  z-index: -3;

  #page {
    z-index: -2;
  }

  #crm-main-content-wrapper {
    position: relative;
    z-index: 0;
  }

  h3,
  h3 + .crm-block {
    position: relative;

    &::before {
      border-radius: $box-radius;
      bottom: 0;
      box-shadow: $box-shadow;
      content: '';
      left: 0;
      position: absolute;
      right: 0;
      top: 0;
      z-index: -1;
    }
  }
  ...
}

```

# Testing

Manual testing + BackstopJS.

PR #404, opened 2019-08-12
SHOR-141: Fix misc issues with CiviCRM and Shoreditch UI

igorpavlov:
# Overview

This PR fixes miscellaneous issues with CiviCRM / Shoreditch styling. Please see the "Changes" section for more information.

# Changes

| What | Before  | After | Any global styling? |
| ------------- | ------------- | ------------- | ------------- |
| Report Register header shadows and action buttons | image | image | Yes, shadows |
| Fix paddings, labels positioning and button icons in Case Search | AwesomeScreenshot-localhost--civicrm-case-search-2019-08-15_3_22 | AwesomeScreenshot-localhost--civicrm-case-search-2019-08-15_3_18 | Yes, icons in buttons image |
| Add a padding on the bottom of form modals | image | image | Yes, applied globally to all modals |
| Improve styling of date filters in Contact Activity | image | image | |
| Remove icons in the contact search selectors | image |  image | Yes, applied globally |

# Technical Overview

## Applying shadows globally

As per https://github.com/civicrm/org.civicrm.shoreditch/pull/402 we applied shadows to the `h3` + `.crm-block` pair (please see the PR for more info). In this PR we extract these styles to apply them globally. It also seemed that some styles from https://github.com/civicrm/org.civicrm.shoreditch/pull/402 were not actually needed. This was thoroughly tested.

```css
.page-civicrm-civirule {
  position: relative;
  z-index: -3;

  #page {
    z-index: -2;
```

^ this has been removed.

## Case Search

This page looks totally different in Vanilla Civi and this is because we destruct the table into blocks:

```css
.crm-accordion-body .form-layout {
  display: block;
  ...

  > tbody:first-child {
    display: block;

    > tr {
      display: block;

      > td {
        display: block;
        ...

        > label:first-child {
          display: block;
```
^ This is a requirement.

## Other changes

Changes other than the ones described above are too trivial, hopefully are self-explanatory and may not worth description.

# Comments

We also now ignore `.DS_Store` files in git.

# Tests

- Manual
- BackstopJS

### DIFF
--- a/scss/civicrm/cases/_search.scss
+++ b/scss/civicrm/cases/_search.scss
@@ -2,73 +2,74 @@
 
 #{civi-page('case-search')} {
   .crm-accordion-body .form-layout {
+    display: block;
     margin-top: $crm-standard-gap;
 
-    &,
-    tbody,
-    tr,
-    td {
+    > tbody:first-child {
       display: block;
-    }
 
-    tbody:first-child > tr {
-      td {
-        margin-left: $crm-standard-gap;
+      > tr {
+        display: block;
 
-        // Display labels as blocks for margins and paddings consistency
-        > label:first-child {
-          display: block;
-
-          // Remove <br>s after labels because labels are displayed as blocks
-          & + br {
-            display: none;
-          }
-        }
-      }
-
-      &:nth-of-type(1) {
         > td {
-          // Brings "Client Name or Email" and the input to individual lines
-          &:nth-of-type(1) {
-            label {
-              display: block;
+          display: block;
+          margin-left: $crm-standard-gap;
 
-              & + input {
-                margin-left: -0.6em;
-              }
+          // Display labels as blocks for margins and paddings consistency
+          > label:first-child {
+            display: block;
+
+            // Remove <br>s after labels because labels are displayed as blocks
+            + br {
+              display: none;
             }
           }
+        }
 
-          // Remove duplicate Search button
-          &:nth-of-type(2) {
-            display: none !important;
+        &:nth-of-type(1) {
+          > td {
+            // Brings "Client Name or Email" and the input to individual lines
+            &:nth-of-type(1) {
+              label {
+                display: block;
+
+                + input {
+                  margin-left: -0.6em;
+                }
+              }
+            }
+
+            // Remove duplicate Search button
+            &:nth-of-type(2) {
+              display: none !important;
+            }
           }
         }
-      }
 
-      // Discard paddings between dates labels and inputs
-      &:nth-of-type(3),
-      &:nth-of-type(4) {
-        > td {
-          &:first-child {
-            padding-bottom: 0;
-          }
+        // Discard paddings between dates labels and inputs
+        &:nth-of-type(3),
+        &:nth-of-type(4) {
+          > td {
+            &:first-child {
+              padding-bottom: 0;
+            }
 
-          &:last-child {
-            padding-top: 0;
+            &:last-child {
+              padding-top: 0;
+            }
           }
         }
-      }
 
-      // Adds a gap between "Case Status" and "Search All Cases / Only My Cases"
-      #s2id_case_status_id {
-        margin-bottom: $crm-standard-gap;
-      }
+        // Adds a gap between "Case Status" and "Search All Cases / Only My Cases"
+        #s2id_case_status_id {
+          margin-bottom: $crm-standard-gap;
+        }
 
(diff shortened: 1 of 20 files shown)


scss/civicrm/cases/_search.scss
```diff
@@ -0,0 +1,76 @@
...
+
+        &:nth-of-type(1) {
+          > td {
+            // Brings "Client Name or Email" and the input to individual lines
+            &:nth-of-type(1) {
+              label {
+                display: block;
+
+                + input {
+                  margin-left: -0.6em;
```
reneolivo: why `0.6em`? do we have a var for this?

scss/civicrm/common/_block-shadows.scss
```diff
@@ -0,0 +1,55 @@
...
+        top: 0;
+        // This is needed to place the shadows underneath so they do not overlap
+        z-index: -1;
+      }
+    }
+  }
+}
+
+h3::before {
+  border-radius: #{$border-radius-base} #{$border-radius-base} 0 0 !important;
```
reneolivo: Can't this really be avoided? Also, you are targeting every single `h3` element. Is this safe?

scss/civicrm/common/_block-shadows.scss
```diff
@@ -0,0 +1,55 @@
...
+
+h3 + .crm-form-block {
+  &,
+  &::before {
+    border-radius: 0 0 #{$border-radius-base} #{$border-radius-base} !important;
+  }
+
+  .crm-submit-buttons {
+    // Need to discard the background otherwise this block will have sharp edges
+    background-color: transparent !important;
```
reneolivo: are all these rules defined with `!important` in CiviCRM?

scss/civicrm/common/_block-shadows.scss
```diff
@@ -0,0 +1,55 @@
+/* stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type, selector-max-id */
```
reneolivo: just as a comment. It seems we keep disabling the same rules again and again. We should either:

* Give up, because we don't have control of the CSS/HTML structure and disable these rules globally.
* Use a `.stylelintrc` file with these rules disabled and place these files in a specific folder within the project.
* Fix the structure of CiviCRM directly (not realistic)

scss/civicrm/common/_forms.scss
```diff
@@ -225,7 +225,7 @@ input.crm-form-password {
   font-family: $font-family-base;
   font-size: $font-size-base;
   height: $input-height-small;
-  padding: 2px 10px;
+  padding: $crm-input-padding;
```
reneolivo: <3

scss/civicrm/cases/_search.scss
igorpavlov: We ar discarding `&nbsp;` here that are hardcoded by Civi. I can leave a comment in SCSS. Would that help?

scss/civicrm/common/_block-shadows.scss
igorpavlov: Seems to be safe @reneolivo. Backstop hasn't found anything. The thing is that H3's are used in so many ways and finally are rendered as block headers. Please note, this only deals with shadows so it is fairly safe.

scss/civicrm/common/_block-shadows.scss
igorpavlov: If we don't use !important we will have to write a completely unreadable chain of elements, plus I want to be 100% sure this exact block never has a background.

scss/civicrm/common/_block-shadows.scss
igorpavlov: This is a requirement in Shoreditch: https://github.com/civicrm/org.civicrm.shoreditch/blob/master/CODING.md

I totally agree @reneolivo this does not look right, but we use linter in Bootstrap.

scss/civicrm/common/_forms.scss
igorpavlov: 😘

scss/civicrm/cases/_search.scss
reneolivo: yes please.

scss/civicrm/common/_block-shadows.scss
reneolivo: If all the files need that rule at the top we can have a `.stylelintrc` file for the `civicrm/` folder only. That will apply the rule only to files under that folder.

Again, nothing to do here. Just a comment.

PR #411, opened 2019-08-22
SHOR-152: Misc issues

igorpavlov:
# Overview

This PR fixes miscellaneous issues with Shoreditch. Please see the **Changes** section below for more info.

# Changes

| What | Before  | After | Any global styling? |
| ------------- | ------------- | ------------- | ------------- |
| Mailing component - Accordion body padding | image | image | Yes |
| Dedupe exceptions - buttons are not styles | image |image | Yes |
| New group modal with custom fields - paddings | image | image| Yes |
| New price field - table paddings | image | image | No |
| New contribution - paddings and unnecessary label |image | image | No |
| Dedupe Exceptions - shadows | image | image | Yes |
| New contact - label font weights | image | image | No |
| Contributions Dashboard - Table paddings | image |image | No |
| New contact - postcode lookup (extension) - input field | image | image | No |
| New contact - email signature | image | image | No |

# Technical Details

While most of the changes are trivial, this one may require attention:

## Styling the rest of buttons

Most CiviCRM buttons are styled by their distinct classes or containers, like here:

```css
&.crm-form-submit:not(.cancel) {
  @extend %btn-civi-primary;
}
```
However, there are some buttons that do not have these classes and are not even in the needed containers. For those we create this style:

```css
div.crm-button:not([class*='crm-button-type']) {
  @extend %btn-civi-primary;
  text-shadow: none;
}
```
We target only `div`s (for `span`s we will then have double styling because `span`s are nesting other `.crm-button` divs inside). We also do not target those which have explicit "types" via `crm-button-type-*` classes.

## Removing inconsistent shadow style

We now use the first shadow in favor of second (see screenshot below): 

image

## Tests

Manual + BackstopJS.

### DIFF
--- a/scss/civicrm/_variables.scss
+++ b/scss/civicrm/_variables.scss
@@ -27,6 +27,9 @@ $crm-light-gray-color: #818c97;
 
 $font-family-fontawesome: 'FontAwesome';
 
+$crm-font-weight-normal: 400;
+$crm-font-weight-bold: 600;
+
 $crm-min-width: 1005px;
 $crm-max-width: 1504px;
 $crm-min-lg-width: 1505px;
--- a/scss/civicrm/_variables.scss
+++ b/scss/civicrm/_variables.scss
@@ -10,7 +10,6 @@ $crm-main-menu-padding-base: 13px;
 $crm-main-menu-padding-small: 8px;
 
 $crm-checkbox-shadow: inset 0 0 3px 0 rgba(0, 0, 0, 0.2);
-$crm-contribute-form-shadow: 0 0 16px 1px rgba(0, 0, 0, 0.1);
 $crm-form-item-shadow: 0 3px 6px 0 rgba(48, 40, 40, 0.25);
 $crm-form-layout-shadow: 0 3px 18px 0 rgba(48, 40, 40, 0.25);
 $crm-form-layout-shadow-lower: 0 8px 18px 0 rgba(48, 40, 40, 0.25);
--- a/scss/civicrm/_variables.scss
+++ b/scss/civicrm/_variables.scss
@@ -9,6 +9,7 @@ $crm-control-height: 30px;
 $crm-main-menu-padding-base: 13px;
 $crm-main-menu-padding-small: 8px;
 
+$crm-box-shadow-light: 0 0 16px 1px rgba(0, 0, 0, 0.1);
 $crm-checkbox-shadow: inset 0 0 3px 0 rgba(0, 0, 0, 0.2);
 $crm-form-item-shadow: 0 3px 6px 0 rgba(48, 40, 40, 0.25);
 $crm-form-layout-shadow: 0 3px 18px 0 rgba(48, 40, 40, 0.25);
--- a/scss/civicrm/common/_base.scss
+++ b/scss/civicrm/common/_base.scss
@@ -249,14 +249,14 @@
   .font-size12pt {
     color: $gray-darker;
     font-size: $font-size-h2;
-    font-weight: 600;
+    font-weight: $crm-font-weight-bold;
     padding-bottom: 10px;
   }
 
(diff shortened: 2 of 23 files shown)


scss/civicrm/_variables.scss
```diff
@@ -27,6 +27,9 @@ $crm-light-gray-color: #818c97;
 
 $font-family-fontawesome: 'FontAwesome';
 
+$crm-font-weight-normal: 400;
+$crm-font-weight-bold: 600;
```
swastikpareek: Bold weight is equivalent of 700.
Check https://htmldog.com/references/css/properties/font-weight/#fbol

scss/civicrm/common/_base.scss
```diff
@@ -149,6 +149,12 @@
     box-shadow: $box-shadow;
     margin-bottom: 20px;
 
+    // Sometimes there are .dataTables_wrapper > .dataTables_wrapper
+    // blocks that produce double shadows
```
swastikpareek: Add `.` in the end.

swastikpareek: Please check this at other places in the css code.

scss/civicrm/contact/pages/_new-individual.scss
```diff
@@ -248,4 +248,30 @@
...
+  .email-signature > .collapsible-title {
+    font-size: $font-size-base;
+    font-weight: $crm-font-weight-bold;
+  }
+
+  // Styling Postcode Lookup extension input
+  #addressLookup input {
+    @include input-text-bootstrap();
+    // Makes the placeholder fully visible
+    min-width: 220px;
```
swastikpareek: Do you think we can document this in a variable?

scss/civicrm/common/_buttons.scss
```diff
@@ -408,3 +408,15 @@ button {
...
   }
 }
+
+// We style only plain buttons which are <div>s/<button>s with `crm-button` class
+// and do *not* have `crm-button-type-*` class. <span>s should not be touched.
+// We do *not* extend to `%btn-civi` because we want the height to be preserved.
+// Because of that we need to additionally discard the text shadow though.
+div,
+button {
+  &.crm-button:not([class*='crm-button-type']) {
```
swastikpareek: Shouldn't this be 
```
&.crm-button:not([class*='crm-button-type-']) {
```
rather for better understanding?

scss/civicrm/common/_modals.scss
```diff
@@ -163,4 +163,8 @@
   .dataTables_wrapper {
     box-shadow: none;
   }
+
+  .crm-form-block h3 {
```
swastikpareek: Do you think we should nest it ?
```
.crm-form-block {
  h3 {
  }
}
```

scss/civicrm/_variables.scss
igorpavlov: This is just a name of variable, we have "bold" font set as 600. Or would you have any suggestions for the variable name?

scss/civicrm/common/_base.scss
igorpavlov: Done.

scss/civicrm/contact/pages/_new-individual.scss
igorpavlov: That would be a bit confusing, because this number just matches the placeholder text length inside the input, so it won't be reused.

scss/civicrm/common/_buttons.scss
igorpavlov: Done

scss/civicrm/common/_modals.scss
igorpavlov: No, why? If there is no other elements, why make it nested?

PR #419, opened 2019-09-09
SHOR-155: Enable Shoreditch As Backend Theme On Install

vinuvarshith:
### Overview
We need to shoreditch theme to be enabled as a theme on installing the extension,  following the [theme system](https://docs.civicrm.org/dev/en/latest/framework/theme/) introduced on CiviCRM v5.16.

### Before
The theme has to be set by going to civicrm/admin/setting/preferences/display?reset=1

### After 
gif40

### Details
In hook_civicrm_install, "custom css url" is unset (if relevant) and 'theme_backend' setting is set to 'shoreditch'. This is done following the upgrader we already have in place [here](https://github.com/civicrm/org.civicrm.shoreditch/blob/d9953d0aee35ab6cf28a53d07ebcead3c427b7cc/CRM/Shoreditch/Upgrader.php#L13-L28).

### DIFF
--- a/shoreditch.php
+++ b/shoreditch.php
@@ -28,6 +28,20 @@ function shoreditch_civicrm_xmlMenu(&$files) {
  */
 function shoreditch_civicrm_install() {
   _shoreditch_civix_civicrm_install();
+
+  // Unset "custom css url" (customCSSURL) field
+  $customCSSUrl = Civi::settings()->get("customCSSURL");
+  if (strpos($customCSSUrl, 'org.civicrm.shoreditch') !== FALSE) {
+    civicrm_api3('setting', 'create', [
+      'customCSSURL' => ''
+    ]);
+  }
+
+  // Set "shoreditch" as the backend theme.
+  civicrm_api3('setting', 'create', [
+    'theme_frontend' => 'default',
+    'theme_backend' => 'shoreditch'
+  ]);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


jamienovick:
Hi @vinuvarshith in the screenshot it still shows the "you have database updates" popup. Why is this? thanks

shoreditch.php
```diff
@@ -28,6 +28,20 @@ function shoreditch_civicrm_xmlMenu(&$files) {
...
+      'customCSSURL' => ''
+    ]);
+  }
+
+  // Set "shoreditch" as the backend theme.
+  civicrm_api3('setting', 'create', [
+    'theme_frontend' => 'default',
+    'theme_backend' => 'shoreditch'
+  ]);
 }
```
davialexandre: @vinuvarshith to avoid the duplication of this setup code, I suggest this:

- Inside `CRM_Shoreditch_Upgrader`, create a `setShoreditchAsBackendTheme()` method and then move all this code there
- Change `upgrade_0001` to simply call this new method instead
- Add an `install()` method to `CRM_Shoreditch_Upgrader` (this will be automatically called during the installation) and call the new method from it as well.

vinuvarshith:
> Hi @vinuvarshith in the screenshot it still shows the "you have database updates" popup. Why is this? thanks

This was an issue. Its now fixed by adding post install hook.
gif41

shoreditch.php
vinuvarshith: done

Issue #425, opened 2019-10-07
Critical: Cannot Edit Existing Navigation Menu


sleewok:
When trying to customize the navigation menu you cannot right click and edit existing items. 

This is pretty important loss of functionality.

T**o fix, the z-index needs to be set to 1 for:**

.jstree-contextmenu.vakata-context

jamienovick:
Please let us know which version of Shoreditch and CiviCRM and base CMS. Thanks

sleewok:
CiviCRM version: 5.18.1
Shoreditch version: Clone of master on github

sleewok:
This is sort of a big deal. Any idea when it will be fixed?

tapashdatta:
I am still having this issue. no context menu loading for navigation. I am on 5.20.3, drupal 7.x
I can see some js error in chrome console

```
js_YL4xAxlPEepJ_TGj59RUxAQcWVrjjCXe9U1tKFAQk6c.js:23 Uncaught TypeError: $(...).afterResize is not a function
    at stickymenu (js_YL4xAxlPEepJ_TGj59RUxAQcWVrjjCXe9U1tKFAQk6c.js:23)
    at HTMLDocument.<anonymous> (js_YL4xAxlPEepJ_TGj59RUxAQcWVrjjCXe9U1tKFAQk6c.js:108)
    at c (js_EebRuRXFlkaf356V0T2K_8cnUVfCKesNTxdvvPSEhCM.js:4)
    at Object.fireWith [as resolveWith] (js_EebRuRXFlkaf356V0T2K_8cnUVfCKesNTxdvvPSEhCM.js:4)
    at Function.ready (js_EebRuRXFlkaf356V0T2K_8cnUVfCKesNTxdvvPSEhCM.js:4)
    at HTMLDocument.q (js_EebRuRXFlkaf356V0T2K_8cnUVfCKesNTxdvvPSEhCM.js:4)
```

sleewok:
Fixed: https://github.com/civicrm/org.civicrm.shoreditch/pull/430

Fabian-SYSTOPIA:
We are still experiencing this issue in CiviCRM 5.29.1 & Shoreditch 0.1-alpha35. We are using the Shoreditch CiviAdmin Companion Drupal theme 1.0.

tschuettler:
> 
> 
[...]

PR #430, opened 2020-01-06
SE-12: Fix edit menu opening issue on nav links

swastikpareek:
## Overview
As a part of this PR the right click triggered menu is now appearing on administration navigation menu link.

## Before
SE-12-before

## After
SE-12-after

## Technical Details
The sub menu was JS behaviour was intact, it was just that it's z-index was not set to the correct value
SE-12-td

## Tests
Manual + BackstopJS.

### DIFF
--- a/scss/civicrm/common/_jstree.scss
+++ b/scss/civicrm/common/_jstree.scss
@@ -1,3 +1,5 @@
+/* stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type */
+
 .jstree-default {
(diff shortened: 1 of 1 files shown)


tapashdatta:
Thanks, but issue still remains. I have just tried the master branch...

sleewok:
This fixes: https://github.com/civicrm/org.civicrm.shoreditch/issues/425

sleewok:
This fixes the following issue: https://github.com/civicrm/org.civicrm.shoreditch/issues/425

PR #432, opened 2020-01-07
SE-13: Fix contribution panel spacing on contacts contribution t…

swastikpareek:
## Overview 
This PR separates the panel spacing between soft credits and normal contributions on the contact contribution pane

## Before
Screenshot 2020-01-07 at 7 01 53 PM

## After
Screenshot 2020-01-07 at 5 23 49 PM

## Technical Details
Top padding is removed from the contribute box and bottom margin is applied and some spacing is adjusted for the soft credit panel too
The table cells of soft credit table are given equal height so that they look similar to contribution table

## Comments
Since the boxes are wrapped with `crm-block` class elements, all those elements had a small border radius and box shadow and hence touching and merging panels to each other was quite a complex and risky solution (considering the implication of the css changes)
So, to keep the UX good and give the impression that these both panes are different, a slight space between the panes of same category is kept. This is done in order to not overlap the box shadow of one pane to another , also if we merge the two panes, due to border radius it won't look good
Screenshot 2020-01-07 at 7 21 09 PM
## Tests
Manual + BackstopJS.

### DIFF
--- a/scss/civicrm/contact/_detail.scss
+++ b/scss/civicrm/contact/_detail.scss
@@ -12,7 +12,7 @@
   }
 
   div.crm-contact-contribute-contributions {
-    padding-top: 20px !important;
+    margin-bottom: 20px;
   }
 
   .crm-summary-block {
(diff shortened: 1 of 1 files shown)


deb1990:
2020-01-08 at 4 59 PM
@swastikpareek can we try to remove these spaces? I think it will be look much better. cc @jamienovick

swastikpareek:
> 2020-01-08 at 4 59 PM
> @swastikpareek can we try to remove these spaces? I think it will be look much better. cc @jamienovick

@deb1990 : This is already been approved by @shitij. As I mentioned in the PR the important motive of this change is to segregate the two panes and moving these panes too close will create issues. Please check the attached gif in technical details
cc @jamienovick

deb1990:
@swastikpareek Yep I saw that later, and removed my comment. All good 👍

PR #444, opened 2020-04-06
CPS-51: Fix dashboard configure buttons padding

reneolivo:
## Overview
This PR fixes the padding for the "done" button located at the dashboard configuration screen.

## Before
Screen Shot 2020-04-06 at 11 30 15 AM

## After
Screen Shot 2020-04-06 at 11 32 43 AM

## Technical details

The padding for this button was not defined correctly when introduced in this PR:https://github.com/civicrm/org.civicrm.shoreditch/pull/48

Even when at the same point in history the main dashboard button padding was defined as `20px 0` instead of `0 20px`:
https://github.com/mukeshcompucorp/org.civicrm.shoreditch/blob/d49e91f0bb3a2c9f164951c3dec52dec71dbaecb/scss/civicrm/dashboard/_dashboard.scss#L2
Change:
https://github.com/civicrm/org.civicrm.shoreditch/pull/48/commits/ffed93838cf0a7f3985b6e9701f1d936b830fcd3#diff-da331dc330bf5fccb6609e55bb475e3dR4

There is no explanation in the PR about this change and why it was introduced. Also, the PR is not even related to the dashboard. This seems to have been a mistake.

We fixed it by using the same value as the main dashboard page, and also replaced the direct `20px` value with a `$crm-standard-gap` variable reference.

Backstop tests from https://github.com/compucorp/backstopjs-config were executed to confirm no regressions were introduced.


deb1990:
This is moved to https://github.com/civicrm/org.civicrm.shoreditch/pull/448

PR #447, opened 2020-04-22
SE-31: Release v0.1 alpha37

deb1990:
Release Update - 22nd April, 2020

### Fixes
1. Fixed a bug which caused broken header in Group Contacts screen.
2. Fixed a bug which caused extra padding in Send Email screen.
3. Fixed a bug which caused styling issues in non modal contribution page table with completed and pending items.
4. Fixed a bug which caused wrong background color of the popups that are opened using non civicrm angular service (dialogService) from core angular.
5. Fixed a bug which caused an extra calendar icon that was being shown as a placeholder for calendar inputs.
6. Fixed a bug which caused spacing issue on assignee contact element on activity form and custom fields option value column alignment.


deb1990:
Branch name was wrong, moved to https://github.com/civicrm/org.civicrm.shoreditch/pull/450

PR #462, opened 2020-07-17
package.json: allow higher nodejs versions

mlutfy:
NodeJS v10 is quite old. This change makes it possible to use a more recent version. I tested with v12 LTS.


deb1990:
This has been fixed in https://github.com/civicrm/org.civicrm.shoreditch/pull/505. And will be released in next version. But master branch can be used for now.

Issue #464, opened 2020-08-07
No access to addtional option on Custom data fields


chumkui:
CiviCRM 5.27.4
Drupal 7.72
Shoreditch 1.0.0-beta.1
Google Chrome

I was trying to delete a custom field today and relalised that there was no option for deletion - all you can do is edit and preview.  I scratched my head a bit and realised that there was no access to the 'more' options on the menu.  All I did was temporarily disable the Shoreditch theme to get the delete option back. I have attached two pngs to describe the issue.
<img width="552" alt="with Shoreditch" src="https://user-images.githubusercontent.com/18424420/89659446-d1501780-d8c7-11ea-8df3-930d4820eb07.png">
<img width="416" alt="without Shoreditch" src="https://user-images.githubusercontent.com/18424420/89659447-d2814480-d8c7-11ea-88f9-7f6bb10b4349.png">

MegaphoneJon:
Duplicates #457.

marcelocompucorp:
#469

marcelocompucorp:
Hi all,

A new Shoreditch version with a fix for this issue has just been released: https://github.com/civicrm/org.civicrm.shoreditch/releases/tag/1.0.0-beta.2

Cheers,

reneolivo:
Closed since this issue has been fixed by https://github.com/civicrm/org.civicrm.shoreditch/pull/469 which has also been released.

Thanks for your help @chumkui

Issue #457, opened 2020-07-10
"More" menu is hidden on "Scheduled Jobs" screen


MegaphoneJon:
This is with the master branch - if you visit **Administer menu » System Settings » Scheduled Jobs**, the **More** menu next to each job is missing, preventing use of "Execute Now" and similar.

MegaphoneJon:
The "More" menu is actually hidden on a lot of pages, like "Manage Groups".  I'm running Shoreditch b1, Civi 5.27.1, with the Shoreditch CiviAdmin Companion Drupal theme 1.0 .

MegaphoneJon:
I did a `git bisect` and traced it back to https://github.com/civicrm/civicrm-core/pull/17295.

jmcclelland:
I'm seeing the same behavior.

MegaphoneJon:
@jmcclelland I haven't had a chance to test - but it looks like #461 might be a solution?

jmcclelland:
It only took me an hour to get gulp to work properly... but no, that didn't solve it for me. 

But I may not be doing it right.

After I apply those changes, the css attributes it removes, remain with the same settings. When I mess with those attributes in inspector I can get the more link to show on the groups page.

I think it's closely related to #461 but how this is handled on the event and contribution pages is different then how it's handled on the manage group (and scheduled jobs)

MegaphoneJon:
@jmcclelland One of my staff worked on this last week, I thought #461 was a better solution if it worked - but since it didn't, here's a change to core.  I'm not sure if it's been tested on non-Shoreditch installs - been meaning to test it myself but haven't had a moment:

```diff
diff --git a/sites/all/modules/civicrm/css/civicrm.css b/sites/all/modules/civicrm/css/civicrm.css
index d260ee66..e0a4bdc6 100644
--- a/sites/all/modules/civicrm/css/civicrm.css
+++ b/sites/all/modules/civicrm/css/civicrm.css
@@ -1555,6 +1555,7 @@ input.crm-form-entityref {
   font-family: "FontAwesome";
   content: "\f0da";
   padding-left: .5ex;
+  display: block;
 }
 
 .crm-container .btn-slide-active .panel {
```

jmcclelland:
Thanks @MegaphoneJon - that fix does work for me. I don't fully understand what's going on with all the css (I see the "..." button, but not "more" and not the nice font awesome triangle promited in the git commit you found). But at least it is functional!

[...]

PR #469, opened 2020-09-03
DRU-256: Fix actions menu icon visibility

ivan-compucorp:
## Overview
There was a bug with **More** menu item of actions menu on most pages - it was not visible (but was present on the page) so it was not possible to click it and select some action. Now its fixed.

## Before
image

## After
image

## Technical Details
To fix the issue we set `position: absolute` to the icon (`::after` element), which was missing and is required to make `right: -2px` work. Also `text-align: center` is added to improve icon alignment inside container.

I've ran backstop tests to verify changes implemented. Some tests failed, but all of them are either expected (because of implemented changes) or false alarms. So we can go ahead with this.
P.S. If you would want to run backstop tests again make sure to add `"misMatchThreshold": 0` to your `backstop.tpl.json` file, otherwise backstop would not notice this small change.

### DIFF
--- a/scss/civicrm/common/_base.scss
+++ b/scss/civicrm/common/_base.scss
@@ -125,7 +125,9 @@
     &::after {
       @include fa-icon($font-size-base, $fa-var-ellipsis-v, $gray-darker);
       background: none;
+      position: absolute;
       right: -2px;
+      text-align: center;
       text-indent: 0;
       top: 0;
       width: 5px;


reneolivo:
@ivan-compucorp can you please mention if this was a regression? This icon has never been visible with Shoreditch or did they stop working at some point?

ivan-compucorp:
@reneolivo I don't have full background here. Maybe this issue would help you? 
https://github.com/civicrm/org.civicrm.shoreditch/issues/457

marcelocompucorp:
Closes #464

reneolivo (approved):

PR #473, opened 2020-09-29
CPS-322: Support new Button markup

deb1990:
## Overview
Recently in the Civicrm Core(https://github.com/civicrm/civicrm-core/pull/18410), all the `input` type button markups has been changed to `button` elements. This caused a lot of styles to break when using Shoreditch theme. So this PR fixes those UI related style issues.

## Before (Only showing some examples)
2020-10-19 at 3 08 PM

2020-10-19 at 3 09 PM

2020-10-19 at 3 10 PM

## After (Only showing some examples)
2020-10-19 at 3 12 PM

2020-10-19 at 3 12 PM

2020-10-19 at 3 11 PM

## Technical Details
1. Changed css selector from `input` to `button` in
```
scss/civicrm/administer/contribute/_managepremiums.scss
scss/civicrm/common/_buttons.scss
scss/civicrm/contact/pages/_contributions.scss
scss/civicrm/financial/_batchtransaction.scss
scss/civicrm/search/pages/_full-text-search.scss
```

2. Made adjustments to the following files to support new markup
```
// There are no extra parent divs for buttons, so moved the css one level up
scss/bootstrap/overrides/crm/_button.scss 

// no extra css required for .crm-i, as items inside buttons are auto aligned
scss/civicrm/administer/display/_display.scss 

// styles for .crm-i removed, as items inside buttons are auto aligned, and crm-i-button does not exist anymore
scss/civicrm/search/pages/_advanced-search.scss 
```

3. Added styles specific to Case Types page in `scss/civicrm/administer/civi-case/_case-types.scss`. And this code was moved from `scss/civicrm/administer/common/_common.scss `. 
This style was moved to the `common.scss` in https://github.com/civicrm/org.civicrm.shoreditch/pull/49. But now it is causing regressions in other pages. And other pages are working fine without these changes, so made it only for case type page.

## Comments
Backstop JS and manual test suites were run to ensure no regressions are present.

### DIFF
--- a/scss/civicrm/administer/display/_display.scss
+++ b/scss/civicrm/administer/display/_display.scss
@@ -73,13 +73,8 @@
         padding: 0;
       }
 
-      .crm-i {
-        margin: 0 -25px 0 15px !important;
-      }
-
       #ckeditor_config {
-        margin: 0 !important;
-        padding-left: 30px;
+        padding: 8px 12px;
       }
     }
   }
--- a/scss/civicrm/common/_buttons.scss
+++ b/scss/civicrm/common/_buttons.scss
@@ -56,10 +56,11 @@ $button-border-radius: 2px;
     }
   }
 
-  input.crm-form-submit {
+  button.crm-form-submit {
     background: $brand-primary;
     border-color: $brand-primary;
     border-radius: $border-radius-base;
+    box-sizing: content-box;
     font-family: $font-family-base;
     margin-right: 0;
     padding: 8px 12px;
@@ -102,7 +103,7 @@ $button-border-radius: 2px;
       }
     }
 
-    input.crm-form-submit {
+    button.crm-form-submit {
       color: $crm-cancel-color !important;
 
       &:hover {
@@ -151,15 +152,12 @@ $button-border-radius: 2px;
 
       .crm-button-type-cancel,
       .crm-button-type-back {
-        input {
-          border: solid 1px $gray-dark;
-          ///The default theme uses "!important"
-          color: $crm-cancel-color !important;
+        border: solid 1px $gray-dark;
+        color: $crm-cancel-color !important; // The default theme uses "!important"
 
-          &:hover {
-            background: darken($crm-copy, $crm-darken-percentage);
-            color: $crm-white !important;
-          }
+        &:hover {
+          background: darken($crm-copy, $crm-darken-percentage);
+          color: $crm-white !important;
         }
       }
     }
@@ -178,8 +176,8 @@ $button-border-radius: 2px;
     line-height: 35px;
   }
 
-  #crm-submit-buttons input.crm-form-submit,
-  .crm-button input.crm-form-submit,
+  #crm-submit-buttons button.crm-form-submit,
+  button.crm-button.crm-form-submit,
   .crm-hover-button,
   .ui-dialog-buttonset .ui-button,
   a.button,
@@ -254,15 +252,13 @@ $button-border-radius: 2px;
     .crm-i {
       background-color: transparent !important;
       color: $crm-white;
-      padding: 2px 3px;
-      top: 0;
+      padding: 0 2px;
     }
   }
 
   .crm-i {
     line-height: inherit;
     text-shadow: none;
-    vertical-align: middle;
   }
 
   .crm-button-type-refresh {
@@ -381,8 +377,9 @@ button {
   border: 0;
   border-radius: $border-radius-base;
   color: $crm-white;
+  cursor: pointer;
   font-family: $font-family-base;
-  line-height: 1;
+  line-height: $line-height-base;
   padding: 11px 12px;
   text-shadow: none;
   text-transform: uppercase;
--- a/scss/civicrm/contact/pages/_contributions.scss
+++ b/scss/civicrm/contact/pages/_contributions.scss
@@ -1015,7 +1015,7 @@ form.CRM_Contribute_Form_ContributionCharts {
 
     button.crm-form-submit {
       margin-left: 5px;
-      padding: 8px 11px;
+      padding: 4px 10px;
       text-shadow: none;
       vertical-align: middle;
     }
--- a/scss/civicrm/contact/pages/_contributions.scss
+++ b/scss/civicrm/contact/pages/_contributions.scss
@@ -1013,9 +1013,9 @@ form.CRM_Contribute_Form_ContributionCharts {
   .form-layout-compressed {
     margin: 20px 0 0;
 
-    input.crm-form-submit {
+    button.crm-form-submit {
       margin-left: 5px;
-      padding: 4px 10px;
+      padding: 8px 11px;
       text-shadow: none;
       vertical-align: middle;
     }
--- a/scss/civicrm/financial/_batchtransaction.scss
+++ b/scss/civicrm/financial/_batchtransaction.scss
@@ -23,7 +23,7 @@
     }
 
     .form-layout-compressed {
-      input.crm-form-submit {
+      button.crm-form-submit {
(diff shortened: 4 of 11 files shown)


eileenmcnaughton:
Just noting I tried this & it fixed the issue I see on 5.31 at

http://dmaster.local/civicrm/admin/uf/group/update?action=update&id=1&context=group



Screen Shot 2020-10-08 at 8 58 56 AM

vs

Screen Shot 2020-10-08 at 9 00 37 AM

deb1990:
@eileenmcnaughton Thanks for checking. PR should be ready within the next week.

scss/civicrm/administer/display/_display.scss
```diff
@@ -73,13 +73,8 @@
...
       }
 
-      .crm-i {
-        margin: 0 -25px 0 15px !important;
-      }
-
       #ckeditor_config {
-        margin: 0 !important;
-        padding-left: 30px;
+        padding: 8px 12px;
```
reneolivo, line 77: why did we change this code for the editor as well?

scss/civicrm/common/_buttons.scss
```diff
@@ -56,10 +56,11 @@ $button-border-radius: 2px;
     }
   }
 
-  input.crm-form-submit {
+  button.crm-form-submit {
```
reneolivo, line 59: isn't it possible to completely remove the element here and just leave the class part?

reneolivo: this is the same question for every instance of `button.class-name`

scss/civicrm/contact/pages/_contributions.scss
```diff
@@ -1013,9 +1013,9 @@ form.CRM_Contribute_Form_ContributionCharts {
   .form-layout-compressed {
     margin: 20px 0 0;
 
-    input.crm-form-submit {
+    button.crm-form-submit {
       margin-left: 5px;
-      padding: 4px 10px;
+      padding: 8px 11px;
```
reneolivo, line 1018: Just a question here, not an action item, but why do we need these very specific padding values? Why not remove them and let the `.crm-form-submit` class from `_button.scss` do it's job? And why such specific values?

scss/civicrm/financial/_batchtransaction.scss
```diff
@@ -23,7 +23,7 @@
     }
 
     .form-layout-compressed {
-      input.crm-form-submit {
+      button.crm-form-submit {
         margin-left: 5px;
         padding: 4px 10px;
```
reneolivo, line 28: In `_contributions.scss` we changed the padding values from `4px 10px` to `8px 11px`. Why wasn't the same change applied here?

scss/civicrm/search/pages/_full-text-search.scss
```diff
@@ -2,9 +2,10 @@
 
 .CRM_Contact_Form_Search_Custom .crm-search-form-block table.form-layout-compressed {
   td {
-    input.crm-form-submit {
+    button.crm-form-submit {
       @extend .btn-sm;
       margin-left: 20px !important;
+      margin-top: -8px;
```
reneolivo, line 8: are we trying to fix a new issue or an existent one? changing from inputs to button misaligned the element?

reneolivo (changes requested):

reneolivo (changes requested):
@deb1990 also, why did the `package-lock.json` file change?

scss/civicrm/common/_buttons.scss
deb1990: I tried this initially @reneolivo . But it messes with the css selector order. I mean if I remove `button` then some of the styles are not applied because of order.

scss/civicrm/search/pages/_full-text-search.scss
deb1990: Yes the position was changed after changing to button. Because we have a long tree of styles being applied from different themes/css file(reset.css, civicrm.css etc).

But I found a better way of fixing this, that is by adding
```
float: none;
vertical-align: initial;
```

scss/civicrm/administer/display/_display.scss
deb1990: 2020-10-20 at 12 41 PM
Its the CK Editor config button actually, which was also changed.

scss/civicrm/common/_buttons.scss
seamuslee001: Just as a thought bubble as well is it worth while keeping input.crm-form-submit and attaching on button.crm-form-submit as another selector for backwards compatibility?

scss/civicrm/financial/_batchtransaction.scss
deb1990: Good catch. the previous one also should be `4px 10px`. Not sure why I made it like that.

scss/civicrm/contact/pages/_contributions.scss
deb1990: Unfortunately paddings are different in different buttons, and right now there is no standard way to do this just by using classes
2020-10-20 at 1 00 PM

deb1990:
> @deb1990 also, why did the package-lock.json file change?

I never knew how this works, but now I have a answer https://stackoverflow.com/a/50868741.
The package-lock.json is `supposed` to be updated every time you run `npm install`. Otherwise we need to run `npm ci` to download exact versions. So in development its fine to use `npm install` and in production `npm ci` should be used.

scss/civicrm/common/_buttons.scss
reneolivo: that's an interesting point @seamuslee001 , but it needs to be a more involved than that. The new buttons don't need icon alignment, but the old ones do. So we would need to preserve both solutions. I think this would bring an issue in the future because it will be hard to debug why we have these statements that are not targeting "anything" (since we'll probably be checking against latest release).

I think the best approach is to define the target CiviCRM version that Shoreditch will be compatible with. If the user upgraders CiviCRM, they must upgrade Shoreditch too and vice versa.

scss/civicrm/contact/pages/_contributions.scss
reneolivo: Thanks @deb1990 I think this is something we might need to look into the future and find out why they behave this way. A PR to CiviCRM-core might be needed to keep them standard.

reneolivo:
@deb1990 can you please use `nvm use` and then `npm install` ? the `package-lock.json` is not supposed to change when installing local packages. The only reason for changing is when installing using a different Node version than the one enforced by NVM/NPM.

scss/civicrm/common/_buttons.scss
deb1990: @seamuslee001 @reneolivo 
I also thought about it initially. But then I felt that its not worth it and is going to be more expensive to maintain both kind of buttons. And if someone really needs supports for `input` elements, then they can use older version of Shoreditch.

reneolivo (approved):

scss/civicrm/common/_buttons.scss
seamuslee001: Yeh fair point just thought throw out the question :)

eileenmcnaughton:
5.31 drops tomorrow so you probably want to merge this & tag a 5.31 compatible release

deb1990:
@eileenmcnaughton Thanks for the reminder. It will be merged today.

deb1990:
@eileenmcnaughton This is the release which supports to support CiviCRM 5.31 buttons.
https://github.com/civicrm/org.civicrm.shoreditch/releases/tag/1.0.0-beta.4

eileenmcnaughton:
yay

PR #488, opened 2021-02-16
Update shoreditch.civix.php

wintermoor:
Deprecated: Array and string offset access syntax with curly braces is deprecated


wintermoor:
Civix will re-generated the file. Has to be fixed here: https://github.com/totten/civix/issues/205

deb1990:
Fixed in https://github.com/civicrm/org.civicrm.shoreditch/pull/507.

PR #494, opened 2021-03-13
Fix buttons that are showing up with white background on front end

larssandergreen:
The later crm-button class background: none was taking precedence over crm-form-submit class background for submit buttons on contribution and event pages. Now these buttons have background: $brand-primary. This is on 5.35 with Shoreditch beta4.


larssandergreen:
Configuration error on my end

PR #495, opened 2021-03-15
CPS-483: Fix dashboard styles

reneolivo:
## Overview

The CiviCRM dashboard has been refactored into a new Angular component and some of the styles, class names, and HTML elements have been updated as well which breaks our current dashboard styles. The changes were introduced in https://github.com/civicrm/civicrm-core/pull/18971

With this PR we fix the dashboard so it looks as it was before.

## How it looked before the Angular update
dashboard-before (1)
Please ignore the "Dashboard" and "My Activities" tab since those are related to the Civicase extension.

## How it looked after the Angular update
CiviCRM-Home-Site-Install

## How it looks with the new fix
dashboard-after (1)

## Technical Details

* Removed the previous dashboard styles since the selectores where no longer valid. Replaced the file with the right styles.
* Added styles for dashlets in individual files located under `scss/civicrm/dashboard/dashlets/`. This will keep the styles separated and organised.
* The `!important` statement in `scss/civicrm/dashboard/dashlets/_news-feed.scss` can't be avoided because of this parent rule: https://github.com/civicrm/org.civicrm.shoreditch/commit/678c2660de525f48c68edeb702e31907ba0bb888#diff-7b9fd65b10b3ef54950a39e7c9ebf467041d11e3c96d64864c585912f706febeR3 It is not clear why this important was added and it might be too risky to remove it for now.
* We removed inline stylelint rules from individual files under the `scss/civicrm` folder. The right way to implement custom rules for a particular folder is using `.stylelintrc` files. The new stylelint file references the root one so we inherit all the rules, but we disable the particular rules that we want to avoid for CiviCRM styles. We also removed the note from the readme file asking developers to add the inline rules since this is no longer necessary. As a note, we disable these rules because we don't have control over the HTML elements and class names in the main project.

## Backstop tests

The Dashboard changes are the only real change detected by the Backstop tests:

https://github.com/compucorp/backstopjs-config/actions/runs/658128879


scss/civicrm/dashboard/_dashboard.scss
```diff
@@ -0,0 +1,57 @@
...
+
+  .crm-dashlet-content {
+    border-radius: 0 0 $border-radius-small $border-radius-small;
+    padding: 0;
+  }
+
+  .crm-dashlet-header {
+    background: $gray-dark;
+    border-radius: $border-radius-small $border-radius-small 0 0;
+    padding: 17px 10px;
```
deb1990, line 47: Cant we use existing values?

scss/civicrm/dashboard/_dashboard.scss
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  .crm-dashlet-header h3 {
+    border: 0;
+    font-size: $font-size-h3;
+    line-height: 135%;
```
deb1990, line 53: Why this 135% value?

scss/civicrm/dashboard/_dashboard.scss
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  .crm-dashlet-header .crm-i {
+    line-height: 135%;
```
deb1990, line 57: Why this 135% value?

scss/civicrm/dashboard/_dashboard.scss
```diff
@@ -0,0 +1,57 @@
...
+  .crm-hover-button {
+    color: $brand-primary;
+
+    &:hover {
+      color: darken($brand-primary, $crm-darken-percentage);
+    }
+  }
+}
+
+#civicrm-dashboard {
```
deb1990, line 20: Cant we remove `#civicrm-dashboard`? that way the styles will be for dashboard styles directly, means h=they can be used in other pages also if needed.

deb1990 (changes requested):

scss/civicrm/dashboard/dashlets/_getting-started.scss
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  table {
+    border: 0;
+    margin: 7px 0;
+  }
+
+  td {
+    border: 0;
+    padding: 7px 15px;
```
deb1990, line 13: Cant we use existing variables? The margin/padding should be uniform in Shoreditch.

scss/civicrm/dashboard/dashlets/_getting-started.scss
```diff
@@ -0,0 +1,19 @@
+#civicrm-getting-started {
+  #help {
+    margin: 0;
+  }
+
+  table {
+    border: 0;
+    margin: 7px 0;
```
deb1990, line 8: Cant we use existing variables? The margin/padding should be uniform in Shoreditch.

scss/civicrm/dashboard/_dashboard.scss
```diff
@@ -1,5 +1,8 @@
 $dashboard-dashlets-margin: 10px;
 
+@import 'scss/civicrm/dashboard/dashlets/getting-started';
```
deb1990, line 3: Can we use `@import 'scss/civicrm/dashboard/dashlets/**';` instead?

deb1990 (changes requested):

scss/civicrm/dashboard/_dashboard.scss
reneolivo: This is very specific to the dashboard. There is another area that is using it:
`scss/civicrm/common/_widget-dashlet.scss`
https://github.com/civicrm/org.civicrm.shoreditch/commit/751749cd045da5a83ac126fb64ca76130e8a46e1#diff-46aa9e81be5c33d36e54dc21cd049c203726a9f35fdee9d175011dd7dba857dbR6

I suspect this is part of the dashboard and needs to be deleted since the selectors should not work, but it's not clear if this was used for widgets outside of the dashboard. The PR that introduced this change didn't explain the reason for the file and why it's under `common` instead of `dashboard`.

This will be left as something to be investigated in the future.

scss/civicrm/dashboard/_dashboard.scss
reneolivo: This value comes from the core style:
https://github.com/civicrm/civicrm-core/blob/master/css/civicrm.css#L235

Will move it to a variable under CRM.

scss/civicrm/dashboard/_dashboard.scss
reneolivo: same as above

scss/civicrm/dashboard/_dashboard.scss
reneolivo: Unfortunately the designs would break and we would need to rely on `!important` statements because some of the rules that we are trying to override use `#civicrm-dashboard`.

scss/civicrm/dashboard/dashlets/_getting-started.scss
reneolivo: This was the previous value we had and this value is also being used in multiple places without being defined in a variable. The weird thing is that is being used to pad tables, buttons, and ui tabs.

We can use Shoreditch defined variables for padding:
<table>
<tr>
<th>Original</th>
<th>Shoreditch Base</th>
<th>Shoreditch Large</th>
</tr>
<tr>
<td align="center">7px 15px</td>
<td align="center">6px 12px</td>
<td align="center">10px 16px</td>
<tr>
<td valign="top">

</td>
<td valign="top">

</td>
<td valign="top">

</td>
</tr>
</table>

Which one should we choose?

scss/civicrm/dashboard/dashlets/_getting-started.scss
reneolivo: will use the value from above.

scss/civicrm/dashboard/_dashboard.scss
reneolivo: It seems globs only work on the root sass files. I tried to use them here, but they didn't work :( might need to investigate how we can use globs in partials. Will leave this for later.

scss/civicrm/dashboard/dashlets/_getting-started.scss
deb1990: Lets keep it as it is then.

deb1990 (approved):

PR #503, opened 2021-03-30
CPS-378: Fix Case Status labels

deb1990:
## Overview
When the Case Status had a Dark Colored background in the Option Settings, the Case Status label were shown as white text on white background, this PR fixes that problem.

## Before
2021-03-30 at 3 51 PM

## After
2021-03-30 at 3 51 PM

## Technical Details
The Background Color of Case Status label should ideally be same as the color set in Option Settings.
But in `scss/civicrm/common/_tables.scss`, the Background was forced to Color White using `!important`, and that ignored the inline background color styles.
Now `!important` is removed, and individual styles are tables to tables where this White Background needs to be forced.

## Comments
Backstop Tests are done in https://github.com/compucorp/backstopjs-config/actions/runs/700673399

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,21 @@
+## Overview
+_A brief description of the pull request. Try to keep it non-technical._
+
+## Before
+_The current status. Please provide screenshots or gifs ([LICEcap](http://www.cockos.com/licecap/), [SilentCast](https://github.com/colinkeenan/silentcast)) where appropriate._
+
+## After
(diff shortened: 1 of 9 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
+
+### Core overrides
+_If the PR overrides/patches a core file, then please explain here, for each file:_
+
+1. _Which file it is_
+2. _The reason why it was overridden/patched_
+3. _What the override/patch consists of_
+
+## Comments
+_Anything else you would like the reviewer to note_
```
reneolivo, line 21: why not add the backstop block here as well?

reneolivo (approved):

PR #505, opened 2021-03-31
CATL-1660: Fix Select 2 styles

deb1990:
## Overview
When having an item with long name inside the Select 2, the UI used to break. This PR fixes that.

## Before
2021-04-02 at 4 14 PM

2021-04-02 at 4 14 PM

## After
2021-04-02 at 4 12 PM

2021-04-02 at 4 13 PM

## Technical Details
In `scss/bootstrap/overrides/vendors/_select2.scss`, added `height: 100%`, to fix the original issue in https://github.com/compucorp/uk.co.compucorp.civicase/pull/279. Now the code from Civicase is removed, as this height is enough to fix the issue.

In `scss/jquery/overrides/_select2.scss`, added `word-break: break-all`, so that long words break, instead of breaking the UI.

Backstop Test : https://github.com/compucorp/backstopjs-config/actions/runs/711467592

Added SCSS Linter as part of build process.
Updated Node JS version to latest LTS i.e 14.16.0.
Also updated `gulp-sass` to latest.


lisandro-compucorp (approved):
👍

reneolivo (approved):

PR #518, opened 2021-05-25
RSE-1608: Adjust style for export dialog

lisandro-compucorp:
## Overview
This PR fixes the style of the form for saving a set of fields after a contact export action.

## Before
On the `Export Options` tab, choosing "Select fields for export":
image

We are moved to select the required fields. Then, the dialog presented there has the "Save fields " button disabled and almost no visible:
image
even after entering text on the input:
image

## After
The button on the dialog window is visible, presenting an appropriate disabled style:
image
After entering text on the input, the disabled style is removed:
image
And also, a hover effect was added:
image

## Technical Details
Backstop Tests: https://github.com/compucorp/backstopjs-config/actions/runs/906013759

### DIFF
--- a/scss/civicrm/contact/pages/_export.scss
+++ b/scss/civicrm/contact/pages/_export.scss
@@ -148,6 +148,8 @@
 }
 
 .crm-export-ui-save-mapping-dialog {
+  $export-form-label-width: 190px;
+
   &.ui-dialog-content.ui-widget-content {
     padding: 0;
   }
@@ -155,17 +157,17 @@
   .crm-form-block {
     box-shadow: none;
     margin-bottom: 0;
-    padding: 20px;
+    padding: $panel-body-padding;
 
     label {
       display: inline-block;
-      width: 190px;
+      width: $export-form-label-width;
     }
   }
 
   .ui-dialog-buttonpane.ui-widget-content {
     height: auto;
-    padding: $modal-inner-padding ($modal-inner-padding * 2);
+    padding: $modal-inner-padding;
     position: inherit;
     width: auto;
 
--- a/scss/civicrm/contact/pages/_export.scss
+++ b/scss/civicrm/contact/pages/_export.scss
@@ -146,3 +146,35 @@
     }
   }
 }
+
+.crm-export-ui-save-mapping-dialog {
+  &.ui-dialog-content.ui-widget-content {
+    padding: 0;
+  }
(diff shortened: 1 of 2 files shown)


scss/civicrm/contact/pages/_export.scss
```diff
@@ -146,3 +146,29 @@
...
+    border: 0;
+    box-shadow: none;
+    float: left;
+    height: auto;
+    margin-top: 30px;
+    padding-bottom: 0;
+    padding-left: 0;
+    padding-right: 0;
+
+    .crm-button.ng-binding {
```
deb1990, line 166: `ng-binding` is a class added by angularjs. It should not be used.

deb1990:
@lisandro-compucorp Good start, but we need to make it look like other popups.
For example 
2021-05-26 at 12 37 PM

Notice, the padding, the buttons, and the background color for the button container.

deb1990 (changes requested):

scss/civicrm/contact/pages/_export.scss
lisandro-compucorp: Removed (I force-pushed the changes then is all in the same commit)

scss/civicrm/contact/pages/_export.scss
```diff
@@ -146,3 +146,40 @@
...
+
+    label {
+      display: inline-block;
+      width: 190px;
+    }
+  }
+
+  .ui-dialog-buttonpane.ui-widget-content {
+    height: auto;
+    padding: 15px 30px;
```
deb1990, line 168: Please use `$modal-inner-padding` variable.
`padding: $modal-inner-padding ($modal-inner-padding * 2);`

deb1990 (changes requested):

scss/civicrm/contact/pages/_export.scss
lisandro-compucorp: Done: https://github.com/civicrm/org.civicrm.shoreditch/pull/518/commits/1655f81443caf4a9f6a41d3f6ba45479acb9d436

deb1990:
@lisandro-compucorp please run backstop and attach a link to the workflow run. You can check previous prs to see how the report is generated.

deb1990 (approved):

deb1990 (approved):

PR #523, opened 2021-07-14
ASESPRT-405: Break long "Location Type" name strings

deb1990:
## Overview
In the contact summary page, if the Name of the Location Type is too long and it is without any Spaces, it breaks the UI. This PR fixes the same.

## Before
2021-07-14 at 1 38 PM

## After
2021-07-14 at 1 37 PM

## Technical Details
Added `word-break: break-all;` to fix this issue.
Also Ignore Linter rules are added for the `scss/civicrm/contact/_detail.scss` file. As this file overrides core styles, it is necessary to ignore these rules.

## Backstop JS Report
https://github.com/compucorp/backstopjs-config/actions/runs/1029463378

### DIFF
--- a/scss/civicrm/contact/_detail.scss
+++ b/scss/civicrm/contact/_detail.scss
@@ -51,6 +51,7 @@
   div.crm-summary-row {
     div.crm-label {
       font-weight: $crm-font-weight-bold;
+      word-break: break-all;
 
       &,
       a {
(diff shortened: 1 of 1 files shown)


lisandro-compucorp (approved):

jamienovick:
Does this affect all labels or just location types? Can we make it clear in the pr? @deb1990

PR #535, opened 2022-03-16
MAE-633: Regenerate CiviCRM menu if it is empty

lisandro-compucorp:
## Overview
This PR solves the theme problem that appears after clearing the DB Cache

## Before
The problem observed consists of the Shoreditch theme not loading correctly at some intervals, but that situation gets solved after refreshing the page. This is an example of how the interface looks:
Screenshot 2021-12-08 at 13 21 14
The error can be reproduced by executing the Job that clears the DB Cache:
capture_broken
**Note** The job usually runs by a console, not using the "Execute now" in the UI as I did on the demo, then the page refresh that we are seeing is not performed. This causes that the next user to access the site after the cron runs in the background sees the error on the theme.

## After
The error after clearing DB Cache does not appear anymore:
capture_fixed

## Technical Details
The function `_shoreditch_isActive` depends on [Civi::service('themes')->getActiveThemeKey()](https://github.com/civicrm/civicrm-core/blob/5.35.2/Civi/Core/Themes.php#L68) to determine if it is active.
This is, in turn, depending on [DrupalBase method](https://github.com/civicrm/civicrm-core/blob/5.35.2/CRM/Utils/System/DrupalBase.php#L658) for detecting if the path corresponds to a frontend or backend theme:
```php
      $config = \CRM_Core_Config::singleton();
      $settingKey = $config->userSystem->isFrontEndPage() ? 'theme_frontend' : 'theme_backend';
```
Since the first moment after the Db Cache clears the menu is empty, the condition of the mentioned method:
```php
    $path = CRM_Utils_System::currentPath();

    // Get the menu for above URL.
    $item = CRM_Core_Menu::get($path);
    // In case the URL is not a civicrm page (a drupal page) we set the FE theme to TRUE - covering the corner case
    return (empty($item) || !empty($item['is_public']));
```
Returns true (since `empty($item) === TRUE`).

For solving this we have two options:
- We detect if it is frontend using another method, like the one suggested [on this comment](https://github.com/civicrm/civicrm-core/blob/5.35.2/Civi/Core/Themes.php#L70)
- We check if the menu is empty and regenerate it.

In this PR I go for the second option, which is easier and acceptable in performance terms. 

## Backstop JS Report
- https://github.com/compucorp/backstopjs-config/actions/runs/1991524616 Ony 4 fails, which corresponds to delays on page load or order in fields.

### DIFF
--- a/shoreditch.php
+++ b/shoreditch.php
@@ -199,5 +199,13 @@ function shoreditch_civicrm_navigationMenu(&$menu) {
  *   TRUE if Shoreditch is the active theme.
  */
 function _shoreditch_isActive() {
+  // The Job civicrm_api3_job_cleanup can clear the DB cache, causing the
+  // CiviCRM menu to be empty on the first load after the command ran.
+  // Without the menu, the getActiveThemeKey is unable to detect the current
+  // theme. This condition ensures the menu is filled.
+  if (!CRM_Core_DAO::checkTableHasData(CRM_Core_DAO_Menu::getTableName())) {
+    CRM_Core_Menu::store(FALSE);
+  }
+
   return Civi::service('themes')->getActiveThemeKey() === 'shoreditch';
 }


jamienovick:
@lisandro-compucorp 
this issue appears to be happening at other times then just when you clear DB cache. Can you confirm this PR fixes it in other situations also?

lisandro-compucorp:
> @lisandro-compucorp this issue appears to be happening at other times then just when you clear DB cache. Can you confirm this PR fixes it in other situations also?

Yes @jamienovick this solves those problems also. I added an extra Note on the PR description, on "Before" section, explaining why.

shoreditch.php
```diff
@@ -199,5 +199,13 @@ function shoreditch_civicrm_navigationMenu(&$menu) {
  *   TRUE if Shoreditch is the active theme.
  */
 function _shoreditch_isActive() {
+  // The Job civicrm_api3_job_cleanup can clear the DB cache, causing the
+  // CiviCRM menu to be empty on the first load after the command ran.
+  // Without the menu, the getActiveThemeKey is unable to detect the current
+  // theme. This condition ensures the menu is filled.
+  if (!CRM_Core_DAO::checkTableHasData(CRM_Core_DAO_Menu::getTableName())) {
+    CRM_Core_Menu::store(FALSE);
+  }
```
ahed-compucorp, line 208: Glad about figuring out the root case of this issue. Although, the code perfectly solves it, I am not in the fever of doing it here at theme level because it affects all CiviCRM themes hence it is a CiviCRM issue.
Maybe we should move the code to [DrupalBase.php#L662](https://github.com/civicrm/civicrm-core/blob/18c0020ce567f61bf0acaa087827c70669f2662e/CRM/Utils/System/DrupalBase.php#L662) or better to build the menu as part of the `civicrm_api3_job_cleanup`  like in [CRM/Upgrade/Form.php#L783](https://github.com/civicrm/civicrm-core/pull/17126/commits/3af5c3fe5f91800f582527da31b089c3bfbedb7d) - see the line  `CRM_Core_Invoke::rebuildMenuAndCaches(FALSE, TRUE);`

ahed-compucorp (changes requested):

ahed-compucorp (approved):

PR #544, opened 2022-07-22
CATL-2243: Fix quick search result dropdown

olayiwola-compucorp:
## Overview
This PR fixes the issue with missing spacing in the quick search result dropdown, this issue is caused by a change in the dropdown item CSS class in the recent CiviCRM version.

The CSS class used to be `ui-menu-item-wrapper` but now `ui-menu-item-uiMenuItemWrapper`, added support for the two classes to ensure backward compatibility.

## Before
On an older version of CiviCRM, 5.35.2
Screenshot 2022-07-22 at 08 08 40

On a newer version of CiviCRM, 5.39.1
Screenshot 2022-07-22 at 08 07 46

## After
On a newer version of CiviCRM, 5.39.1
Screenshot 2022-07-22 at 09 21 55

## Backstop JS Report
[Backstop JS report here](https://github.com/compucorp/backstopjs-config/actions/runs/2717988065)

### DIFF
--- a/scss/civicrm/common/_main-menu.scss
+++ b/scss/civicrm/common/_main-menu.scss
@@ -108,6 +108,7 @@
     font-size: $font-size-base;
     padding: $crm-main-menu-padding-base 0;
 
+    .ui-menu-item-uiMenuItemWrapper,
     .ui-menu-item-wrapper {
       padding: $crm-main-menu-padding-small ($crm-main-menu-padding-base * 2) $crm-main-menu-padding-small $crm-main-menu-padding-base;
     }


erawat (commented):
@olayiwola-compucorp Since this changes, does it still work with version 5.35.2? If not, then, we would need to change the minimum requirements in the readme and info file to indicate supported version.

olayiwola-compucorp:
Yes @erawat, I mentioned in the PR description that the previously supported version would not be affected.

> added support for the two classes to ensure backward compatibility.

swastikpareek (approved):

PR #569, opened 2025-02-13
COMCL-1029: Dropdown overflow should be visible

olayiwola-compucorp:
## Overview
This PR ensures dropdown overflows are visible

## Before

https://github.com/user-attachments/assets/7679fe26-1171-46e6-a0b2-d1806b89e523


shahrukh-compuco (approved):

jamienovick:
@olayiwola-compucorp this doesn't seem to have fully solved the issue as some of the overflow cannot be seen over the CiviCRM container footer. 

Can we fix that please?

Cc @shahrukh-compuco

PR #571, opened 2025-02-19
Make the New Mailing Subject field 100% of container width

dwoods-encircle:
## Overview
Make the New Mailing Subject field 100% of container width

## Before
It's too small.

## After
It's full width.

## Technical Details
Just one new SCSS file "_mailing-create.scss"


dwoods-encircle:
The CSS hasn't been generated from the new SCSS file. This will need doing.

dwoods-encircle:
Doh, just realised this would only apply to Mosaico Subject field - will cancel this one and submit a new one.

